PR #2, opened 2024-01-10
Load test for writing observations into the datastore

lukas-phaf:
Add load testing for writing observations into datastore using apscheduler.

The code for the first iteration (using locust) is also still there, in the files `locust_write.py` and `netcdf_file_to_requests.py`. But these are not used in the Github Actions.


datastore/load-test/locustfile.py
```diff
@@ -34,6 +34,7 @@
 class StoreGrpcUser(grpc_user.GrpcUser):
     host = "localhost:50050"
     stub_class = dstore_grpc.DatastoreStub
+    weight = 1
```
lukas-phaf, line 37: `locustfile.py` can now be deleted, I think.

datastore/load-test/requirements.txt
```diff
@@ -1,18 +1,21 @@
 #
-# This file is autogenerated by pip-compile with Python 3.11
+# This file is autogenerated by pip-compile with Python 3.10
```
lukas-phaf, line 2: Regenerate with Python 3.11.

datastore/load-test/locustfile_write.py
Jeffrey-Vervoort-KNMI, line 1: Do you want to keep this file? As you created another way of running the performance tests? With a cron scheduler.

datastore/load-test/schedule_write.py
```diff
@@ -0,0 +1,101 @@
...
+from google.protobuf.timestamp_pb2 import Timestamp
+from variables import variable_info
+
+# from apscheduler.executors.pool import ProcessPoolExecutor
+
+
+# One channel & client per process (not per thread!)
+# TODO: Does this affect load on the database? Seems load here is lower then in Rosina's code
+channel = grpc.insecure_channel(f"{os.getenv('DSHOST', 'localhost')}:{os.getenv('DSPORT', '50050')}")
+client = dstore_grpc.DatastoreStub(channel=channel)
```
Jeffrey-Vervoort-KNMI, line 25: Can a single channel be a bottleneck for the load test? I am wondering if this could be the case as I read that a channel has a limit on the number of concurrent streams. When the limit is reached, the RPCs are queued.

https://grpc.io/docs/guides/performance/

datastore/load-test/schedule_write.py
```diff
@@ -0,0 +1,101 @@
...
+    # scheduler.add_executor(ProcessPoolExecutor())
+    scheduler.add_executor(ThreadPoolExecutor())
+    print(f"Now: {datetime.now()}", flush=True)
+    for i in range(0, 5000):
+        (period, cron) = random.choice(crons)
+        station_id = f"station{i:04d}"
+        station = Station(station_id, random.uniform(50.0, 55.0), random.uniform(4.0, 8.0), period)
+        # print(station_id, cron, period)
+        # TODO: Spread less well over time, for example, all use same second, but add jitter < 60
+        trigger = CronTrigger(minute=cron, second=random.randint(0, 59), jitter=1)
```
Jeffrey-Vervoort-KNMI, line 95: Do you think it is necessary to keep the comment and the TODO? As there is some jitter

datastore/load-test/schedule_write.py
Jeffrey-Vervoort-KNMI: Added as a comment, will look at it some time later.

Jeffrey-Vervoort-KNMI (approved):

PR #13, opened 2024-01-12
Fix bug that cleanup runs after every insert.

lukas-phaf:
Cleanup interval defaulted to 86400 nanoseconds instead of seconds.


datastore/datastore/storagebackend/postgresql/postgresql.go
```diff
@@ -31,7 +31,7 @@ func initCleanupInterval() {
 		val = defaultVal
 	}
 
-	cleanupInterval = time.Duration(val)
+	cleanupInterval = time.Duration(val) * time.Second
```
jo-asplin-met-no, line 34: LGTM - 👍

jo-asplin-met-no (approved):

PR #23, opened 2024-01-30
Add remaining acdd attrs issue 54

jo-asplin-met-no:
This branch now [includes](https://github.com/EURODEO/e-soh/commit/8b6054ecdb2eedcf3f09a125446c52f200642729) the suggestion of using a `map<string, Strings> filter` in `GetObsRequest` to save explicit references to field names. Note how the example in the README.md file is updated.


jo-asplin-met-no:
Need to update integration tests. In progress ...

jo-asplin-met-no:
Stuck in the test failures in CI/test-datastore and CI/test-ingest. Does anyone have an idea of what needs to be fixed here?

protobuf/datastore.proto
```diff
@@ -139,14 +151,33 @@ message PutObsResponse {
...
+
+  // --- END non-string metadata -------------------------
+
+  // --- BEGIN string metadata (handleable with reflection) -------------------------
+
+  // general filter
+  // - map keys must correspond exactly with string field names in TSMetadata or ObsMetadata
+  // - if map key F is specified (where F is for example 'platform'), only observations that match
+  //   at least one of these values for F may be returned
+  map<string, Strings> filter = 3;
```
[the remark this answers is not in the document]
lukas-phaf: Hmm, looking at the Python, what you have now is probably best.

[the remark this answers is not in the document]
jo-asplin-met-no: I wish it was possible in Python to write just

```
dstore.GetObsRequest(filter=dict(instrument=["tn", ...]))
```
or
```
dstore.GetObsRequest(filter={'instrument':["tn", ...]})
```
instead of
```
dstore.GetObsRequest(filter=dict(instrument=dstore.Strings(values=["tn"])))
```
But I guess one gets used to the more explicit form,

datastore/datastore/storagebackend/postgresql/putobservations.go
```diff
@@ -104,6 +77,26 @@ func getTSColVals(tsMdata *datastore.TSMetadata) ([]interface{}, error) {
...
+		method := rv.MethodByName(methodName)
+		if field.IsExported() && (field.Type.Kind() == reflect.String) && (method.IsValid()) {
+			val, ok := method.Call([]reflect.Value{})[0].Interface().(string)
+			if !ok {
+				return nil, fmt.Errorf(
+					"method.Call() failed for method %s; failed to return string", methodName)
+			}
+			colVals = append(colVals, val)
+		}
+	}
```
lukas-phaf, line 96: I was wondering if there is a way to do all the reflection upfront, and don't do it at runtime (as we are already somewhat limited on insertion speed... but that might very well not be the go code, but the DB)...

But anyway, if you want to go down the rabbit hole with me...

Go has a thing called _method expression_, with which you can call the method providing the receiver as a argument. So instead of `p.Distance(q)` you write:
```
distance := Point.Distance
distance(p, q)
```

If a _method expression_ could be derived using reflection for each of the metadata fields, we could do that upfront, and then simply call those in a loop.
But I haven't been able to find an example of how to do this. Maybe it is not possible, and if it is, maybe the performance is just as bad.

lukas-phaf: I'll bring over my write load test from the other branch, and take the current code for a spin.

datastore/datastore/storagebackend/postgresql/getobservations.go
```diff
@@ -14,6 +16,103 @@ import (
...
+		pb2table[pbName] = "observation"
+	}
+
+	// create pb2tableKeys
+	pb2tableKeys = []string{}
+	for key := range pb2table {
+		pb2tableKeys = append(pb2tableKeys, key)
+	}
+	sort.Strings(pb2tableKeys)
+}
```
lukas-phaf, line 94: Move this to a seperate file?

datastore/datastore/storagebackend/postgresql/getobservations.go
```diff
@@ -219,100 +331,199 @@ func getGeoFilter(inside *datastore.Polygon, phVals *[]interface{}) (string, err
...
+		},
+		Obstime: &datastore.ObsMetadata_ObstimeInstant{
+			ObstimeInstant: timestamppb.New(obsTimeInstant0),
+		},
+		Pubtime: timestamppb.New(pubTime0),
+		Value:   value,
+	}
+
+	// complete obsMdata with string metadata (handleable with reflection)
+	err := addStringMdata(reflect.ValueOf(&obsMdata), obsStringMdataGoNames, colVals0)
```
lukas-phaf, line 460: This is brilliant!

datastore/datastore/storagebackend/postgresql/getobservations.go
```diff
@@ -219,100 +331,199 @@ func getGeoFilter(inside *datastore.Polygon, phVals *[]interface{}) (string, err
...
 	}
 
 	// get time series
-	tsMdata := map[int64]*datastore.TSMetadata{}
+	tsMdatas := map[int64]*datastore.TSMetadata{}
 	tsIDs := []string{}
-	for id := range obsMap {
-		tsIDs = append(tsIDs, fmt.Sprintf("%d", id))
+	for tsID := range obsMdatas {
+		tsIDs = append(tsIDs, fmt.Sprintf("%d", tsID))
```
lukas-phaf, line 517: I might be missing something... but if we have multiple observations from a time series, won't there by many duplicated in the `tdID`'s? This seem to make it all the way into the SQL query?

Probably easiest to make tsID's a "set"?

datastore/datastore/storagebackend/postgresql/putobservations.go
lukas-phaf: Running my "writer" load test (from the other PR) on your branch, I don't see any problems. It runs fine, not errors in the first hour. I don't see a significant higher load (it might be 10% higher, hard to tell by just watching the numbers.

For the read-only load (I compared Github actions for your and my branch), they are pretty much the same.

May the conclusion is that even if reflection is slower, it is still not relevant compared to the DB performance?

datastore/datastore/storagebackend/postgresql/getobservations.go
jo-asplin-met-no: I do think you missed something. A given tsID is processed only once since the loop iterates over the keys of obsMdatas which form a set by definition (just like the keys of any map).

datastore/datastore/storagebackend/postgresql/getobservations.go
jo-asplin-met-no: That is a good idea. How about introducing an init.go (in the same directory) that takes care of all initialization for the postgresql package? Then we can move all package variables and their initialization code to that init.go. Note that there is no concept of file scope in Go, so all package variables (i.e. starting with lowercase and thus only accessible within the package) might as well be declared in a single file. And we could use the same organization for other packages too.

datastore/datastore/storagebackend/postgresql/putobservations.go
jo-asplin-met-no: I think this should be feasible, and also save a bit of time by doing certain reflection operations once and for all in init(). How much time I'm like you not sure of, but I'll commit a change for it, so we can easily switch between the two implementations if we like.

datastore/datastore/storagebackend/postgresql/getobservations.go
jo-asplin-met-no: Added [a commit](https://github.com/EURODEO/e-soh/pull/23/commits/bcf9a21719cfe879627192a43b7bc27ef4c9c8ed) to the PR just now for this.

datastore/datastore/storagebackend/postgresql/getobservations.go
lukas-phaf: You are right of course. I don't know what my issue was. I'll blame it on it being 6pm.

datastore/datastore/storagebackend/postgresql/putobservations.go
jo-asplin-met-no: Made a change for this now. Though not entirely sure it was what you had in mind since it was a quite straightforward refactoring that didn't really introduce anything new except from moving some reflection calls to init().

lukas-phaf (approved):

PR #25, opened 2024-02-02
Issue 17 proto object to covjson conversion unit tests

### LINKED ISSUE #17: Unit test for protobuf to CoverageJSON conversion
We should generate the protobuf from Python, so that we don't need to put binary message in the repo.

fjugipe:
Closes #17

- Adds unit tests to  protobuf to coverageJson formatter
  - Test cases for different conversions mocking the gRPC response and checking the conversion against known conversion results
- Update Github Actions to run api unit tests on push
  - Post coverage to pull request comment or push commit depending on action

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,9 @@ on:
       - "**"
     tags:
       - "[0-9]+.[0-9]+.[0-9]+"
+  pull_request:
+    branches:
+      - main
 
 permissions:
   contents: read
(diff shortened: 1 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -6,6 +6,9 @@ on:
       - "**"
     tags:
       - "[0-9]+.[0-9]+.[0-9]+"
+  pull_request:
+    branches:
+      - main
```
lukas-phaf, line 11: What does this do, as we already run on every push?

fjugipe: It is needed to comment the Coverage report to pull request. Otherwise it will be posted to push commit.

lukas-phaf (approved):

PR #30, opened 2024-02-08
Issue 18 functionality to return all parameters when parameter-names are not given

### LINKED ISSUE #18: Add funcitonality to return all parameters when parameter_name is not given
Please add labels for relevant API endpoints.

fjugipe:
Closes #18 

- Parameter name is no longer required in the query
  - Returns all parameters and values from /locations/{id}, /position and /area endpoints without parameter name in query
    - Returns all also when parameter-name is given, but empty string
  
- Updates Swagger documentation with examples of time date

- Improves error handling
  - Return 422 with incorrect coordinates or geometric type
  - Return 422 when time date is wrong format or start time is before end time in query

- Adds unit tests to edr api endpoints
  - Test cases with mocked functions asserting correct calls to functions
  - Test cases to check that incorrect queries returns 422 with descriptive error message

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -129,6 +129,7 @@ jobs:
           pip install --upgrade pip
           pip install pytest-timeout
           pip install pytest-cov
+          pip install httpx
           pip install -r ./api/requirements.txt
 
       - name: Copy Protobuf file to api directory and build
--- a/api/dependencies.py
+++ b/api/dependencies.py
@@ -2,6 +2,7 @@ from datetime import datetime
 from datetime import timedelta
 from typing import Tuple
 
+from fastapi import HTTPException
 from google.protobuf.timestamp_pb2 import Timestamp
 from pydantic import AwareDatetime
 from pydantic import TypeAdapter
@@ -11,23 +12,38 @@ def get_datetime_range(datetime_string: str | None) -> Tuple[Timestamp, Timestam
     if not datetime_string:
         return None
 
+    errors = {}
+
     start_datetime, end_datetime = Timestamp(), Timestamp()
     aware_datetime_type_adapter = TypeAdapter(AwareDatetime)
-    datetimes = tuple(value.strip() for value in datetime_string.split("/"))
-    if len(datetimes) == 1:
-        start_datetime.FromDatetime(aware_datetime_type_adapter.validate_python(datetimes[0]))
-        end_datetime.FromDatetime(
-            aware_datetime_type_adapter.validate_python(datetimes[0]) + timedelta(seconds=1)
-        )  # HACK: Add one second so we get some data, as the store returns [start, end)
-    else:
-        if datetimes[0] != "..":
+
+    try:
+        datetimes = tuple(value.strip() for value in datetime_string.split("/"))
+        if len(datetimes) == 1:
             start_datetime.FromDatetime(aware_datetime_type_adapter.validate_python(datetimes[0]))
+            end_datetime.FromDatetime(
+                aware_datetime_type_adapter.validate_python(datetimes[0]) + timedelta(seconds=1)
+            )  # HACK: Add one second so we get some data, as the store returns [start, end)
         else:
-            start_datetime.FromDatetime(datetime.min)
-        if datetimes[1] != "..":
-            # HACK add one second so that the end_datetime is included in the interval.
-            end_datetime.FromDatetime(aware_datetime_type_adapter.validate_python(datetimes[1]) + timedelta(seconds=1))
-        else:
-            end_datetime.FromDatetime(datetime.max)
+            if datetimes[0] != "..":
+                start_datetime.FromDatetime(aware_datetime_type_adapter.validate_python(datetimes[0]))
+            else:
+                start_datetime.FromDatetime(datetime.min)
+            if datetimes[1] != "..":
+                # HACK add one second so that the end_datetime is included in the interval.
+                end_datetime.FromDatetime(
+                    aware_datetime_type_adapter.validate_python(datetimes[1]) + timedelta(seconds=1)
+                )
+            else:
+                end_datetime.FromDatetime(datetime.max)
+
+            if start_datetime.seconds > end_datetime.seconds:
+                errors["datetime"] = f"Invalid range: {datetimes[0]} > {datetimes[1]}"
+
+    except ValueError:
+        errors["datetime"] = f"Invalid format: {datetime_string}"
+
+    if errors:
+        raise HTTPException(status_code=422, detail=errors)
 
     return start_datetime, end_datetime
--- a/api/routers/edr.py
+++ b/api/routers/edr.py
@@ -62,7 +62,7 @@ async def get_locations(bbox: str = Query(..., example="5.0,52.0,6.0,52.1")) -> 
 async def get_data_location_id(
     location_id: str = Path(..., example="06260"),
     parameter_name: str = Query(None, alias="parameter-name", example="dd,ff,rh,pp,tn"),
-    datetime: str | None = None,
+    datetime: str = Query(None, example="2022-12-31T00:00Z/2023-01-01T00:00Z"),
     f: str = Query(default="covjson", alias="f", description="Specify return format."),
 ):
     # TODO: There is no error handling of any kind at the moment!
@@ -90,7 +90,7 @@ async def get_data_location_id(
 async def get_data_position(
     coords: str = Query(..., example="POINT(5.179705 52.0988218)"),
     parameter_name: str = Query(None, alias="parameter-name", example="dd,ff,rh,pp,tn"),
-    datetime: str | None = None,
+    datetime: str = Query(None, example="2022-12-31T00:00Z/2023-01-01T00:00Z"),
     f: str = Query(default="covjson", alias="f", description="Specify return format."),
 ):
     point = wkt.loads(coords)
@@ -108,7 +108,7 @@ async def get_data_position(
 async def get_data_area(
     coords: str = Query(..., example="POLYGON((5.0 52.0, 6.0 52.0,6.0 52.1,5.0 52.1, 5.0 52.0))"),
     parameter_name: str = Query(None, alias="parameter-name", example="dd,ff,rh,pp,tn"),
-    datetime: str | None = None,
+    datetime: str = Query(None, example="2022-12-31T00:00Z/2023-01-01T00:00Z"),
     f: str = Query(default="covjson", alias="f", description="Specify return format."),
 ):
     poly = wkt.loads(coords)
--- a/api/routers/edr.py
+++ b/api/routers/edr.py
@@ -56,12 +56,12 @@ async def get_locations(bbox: str = Query(..., example="5.0,52.0,6.0,52.1")) -> 
 @router.get(
     "/locations/{location_id}",
     tags=["Collection data queries"],
-    response_model=Coverage,
+    response_model=Coverage | CoverageCollection,
     response_model_exclude_none=True,
 )
 async def get_data_location_id(
     location_id: str = Path(..., example="06260"),
-    parameter_name: str = Query(..., alias="parameter-name", example="dd,ff,rh,pp,tn"),
+    parameter_name: str = Query(None, alias="parameter-name", example="dd,ff,rh,pp,tn"),
     datetime: str | None = None,
     f: str = Query(default="covjson", alias="f", description="Specify return format."),
 ):
@@ -71,7 +71,9 @@ async def get_data_location_id(
     get_obs_request = dstore.GetObsRequest(
         filter=dict(
             platform=dstore.Strings(values=[location_id]),
-            instrument=dstore.Strings(values=list(map(str.strip, parameter_name.split(",")))),
+            instrument=dstore.Strings(
+                values=list(map(str.strip, parameter_name.split(","))) if parameter_name else "*"
+            ),
         ),
         interval=dstore.TimeInterval(start=range[0], end=range[1]) if range else None,
     )
--- a/api/routers/edr.py
+++ b/api/routers/edr.py
@@ -5,6 +5,7 @@ from covjson_pydantic.coverage import Coverage
 from covjson_pydantic.coverage import CoverageCollection
 from dependencies import get_datetime_range
 from fastapi import APIRouter
+from fastapi import HTTPException
 from fastapi import Path
 from fastapi import Query
 from geojson_pydantic import Feature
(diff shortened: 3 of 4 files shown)


api/routers/edr.py
```diff
@@ -105,15 +112,21 @@ async def get_data_position(
...
-    datetime: str | None = None,
+    parameter_name: str = Query(None, alias="parameter-name", example="dd,ff,rh,pp,tn"),
+    datetime: str = Query(None, example="2022-12-31T00:00Z/2023-01-01T00:00Z"),
     f: str = Query(default="covjson", alias="f", description="Specify return format."),
 ):
-    poly = wkt.loads(coords)
-    assert poly.geom_type == "Polygon"
+    try:
+        poly = wkt.loads(coords)
+        assert poly.geom_type == "Polygon"
```
Teddy-1000, line 123: I think we should raise different errors for failure to parse the wkt or if the wrong type of geometry is provided.

api/routers/edr.py
```diff
@@ -87,13 +90,17 @@ async def get_data_location_id(
...
+    parameter_name: str = Query(None, alias="parameter-name", example="dd,ff,rh,pp,tn"),
+    datetime: str = Query(None, example="2022-12-31T00:00Z/2023-01-01T00:00Z"),
     f: str = Query(default="covjson", alias="f", description="Specify return format."),
 ):
-    point = wkt.loads(coords)
-    assert point.geom_type == "Point"
-    poly = buffer(point, 0.0001, quad_segs=1)  # Roughly 10 meters around the point
+    try:
+        point = wkt.loads(coords)
+        assert point.geom_type == "Point"
```
Teddy-1000, line 101: Same as in get_data_area, error should differentiate between error in parsing wkt and providing wrong geometry type.

Teddy-1000 (changes requested):
Test looks good, added some comments in dependencies.py

api/routers/edr.py
```diff
@@ -58,13 +59,13 @@ async def get_locations(bbox: str = Query(..., example="5.0,52.0,6.0,52.1")) ->
...
     tags=["Collection data queries"],
-    response_model=Coverage,
+    response_model=Coverage | CoverageCollection,
     response_model_exclude_none=True,
 )
 async def get_data_location_id(
     location_id: str = Path(..., example="06260"),
-    parameter_name: str = Query(..., alias="parameter-name", example="dd,ff,rh,pp,tn"),
-    datetime: str | None = None,
+    parameter_name: str = Query(None, alias="parameter-name", example="dd,ff,rh,pp,tn"),
```
lukas-phaf, line 68: I am not a big fan of this, we are basically saying that the type is `str`, but the default is `None`, which doesn't fit in that type.

Also, providing the default through `Query` like this is now deprecated. How about this?

```suggestion
    parameter_name: Annotated[str | None, Query(alias="parameter-name", example="dd,ff,rh,pp,tn")] = None,
):
```
See https://fastapi.tiangolo.com/tutorial/query-params-str-validations/#add-more-validations

Same in several other places.

api/routers/edr.py
```diff
@@ -73,7 +74,9 @@ async def get_data_location_id(
     get_obs_request = dstore.GetObsRequest(
         filter=dict(
             platform=dstore.Strings(values=[location_id]),
-            instrument=dstore.Strings(values=list(map(str.strip, parameter_name.split(",")))),
+            instrument=dstore.Strings(
+                values=list(map(str.strip, parameter_name.split(","))) if parameter_name else "*"
+            ),
```
lukas-phaf, line 80: Better to leave the whole `instrument` field out if you don't want to filter on it. I think the code will be slightly clearer, and it will more efficient in the store.

Also for the other endpoint.

api/routers/edr.py
lukas-phaf: I don't think we want to handle this with an `assert` in production code.

api/dependencies.py
```diff
@@ -11,23 +12,38 @@ def get_datetime_range(datetime_string: str | None) -> Tuple[Timestamp, Timestam
...
+                end_datetime.FromDatetime(datetime.max)
+
+            if start_datetime.seconds > end_datetime.seconds:
+                errors["datetime"] = f"Invalid range: {datetimes[0]} > {datetimes[1]}"
+
+    except ValueError:
+        errors["datetime"] = f"Invalid format: {datetime_string}"
+
+    if errors:
+        raise HTTPException(status_code=422, detail=errors)
```
lukas-phaf, line 47: I agree 422 is the most appropriate error code... but EDR spec doesn't seem to allow it:
https://docs.ogc.org/is/19-086r6/19-086r6.html#_e006a89e-6e26-4c11-b1e3-15988d48eb14

From the description in that list, I think 400 is appropriate.

api/routers/edr.py
fjugipe: Added explicit error messages to parsing & wrong geom types.

api/routers/edr.py
fjugipe: Added explicit error messages to parsing & wrong geom types.

api/dependencies.py
```diff
@@ -11,23 +12,38 @@ def get_datetime_range(datetime_string: str | None) -> Tuple[Timestamp, Timestam
     if not datetime_string:
         return None
 
+    errors = {}
+
```
lukas-phaf, line 16: As there are only two errors types, it is maybe more straightforward to just throw the exception directly?

api/test/test_edr.py
```diff
@@ -0,0 +1,186 @@
...
+
+def create_mock_obs_response(json_data):
+    response = dstore.GetObsResponse()
+    Parse(json.dumps(json_data), response)
+    return response
+
+
+def load_json(file_path):
+    with open(file_path, "r") as file:
+        return json.load(file)
```
lukas-phaf, line 186: Duplicate with other test file?

api/test/test_edr.py
```diff
@@ -0,0 +1,186 @@
...
+        mock_getObsRequest.return_value = create_mock_obs_response(test_data)
+
+        response = client.get(
+            "/collections/observations/locations/06260?"
+            + "parameter-name=ff&datetime=2022-12-31T00:00:00Z/2022-12-31T01:00:00Z"
+        )
+
+        # Check that getObsRequest gets called with correct arguments given in query
+        mock_getObsRequest.assert_called_once()
+        assert "ff" in mock_getObsRequest.call_args[0][0].filter["instrument"].values
```
lukas-phaf, line 28: Can we check it only has `ff`.

api/test/test_edr.py
```diff
@@ -0,0 +1,186 @@
...
+        assert "06260" in mock_getObsRequest.call_args[0][0].filter["platform"].values
```
lukas-phaf, line 29: And only `620`

.github/workflows/ci.yml
```diff
@@ -129,6 +129,7 @@ jobs:
           pip install --upgrade pip
           pip install pytest-timeout
           pip install pytest-cov
+          pip install httpx
```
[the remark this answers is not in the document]
fjugipe: to use TestClient to test the api we need httpx

api/dependencies.py
fjugipe: My initial thought was to make it expendable. So if we want to apply some other restrictions to time range in the future, we can respond with multiple invalid parameter messages at once and the user can correct all the mistakes at once instead of several requests.

.github/workflows/ci.yml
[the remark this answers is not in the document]
lukas-phaf: Ah, clear:
https://fastapi.tiangolo.com/tutorial/testing/#using-testclient

Always a bit confusing if packages don't manage their own dependencies.

api/routers/edr.py
fjugipe: Changed to raise and catch TypeErrors instead of asserts in https://github.com/EURODEO/e-soh/pull/30/commits/b732cdcc1420137f7882a7c52b10e702eef9781d

api/routers/edr.py
fjugipe: Refactored in https://github.com/EURODEO/e-soh/pull/30/commits/f1b1986323ad0b76c2eac6103f256ad90e201a91

api/dependencies.py
fjugipe: Now returns correct codes as per specification https://github.com/EURODEO/e-soh/pull/30/commits/f700ccb64c55286be9ef60a7aba6bcb612a0ce51

api/routers/edr.py
fjugipe: Now leaves `instrument` out when not needed https://github.com/EURODEO/e-soh/pull/30/commits/36e1a0e93a46b2f9a619f2f29a9eb16278107318

api/test/test_edr.py
fjugipe: Added utilities.py to test folder containing helper functions. https://github.com/EURODEO/e-soh/pull/30/commits/b34ed81668c048efaf6160a9fa8df2daf8eb1597

api/test/test_edr.py
fjugipe: https://github.com/EURODEO/e-soh/pull/30/commits/1ac4ad6138263bf82b8ca83b91571c18145611b4

api/test/test_edr.py
fjugipe: https://github.com/EURODEO/e-soh/pull/30/commits/1ac4ad6138263bf82b8ca83b91571c18145611b4

api/routers/edr.py
```diff
@@ -89,13 +97,24 @@ async def get_data_location_id(
...
-    poly = buffer(point, 0.0001, quad_segs=1)  # Roughly 10 meters around the point
+    try:
+        point = wkt.loads(coords)
+        if point.geom_type != "Point":
+            raise TypeError
+        poly = buffer(point, 0.0001, quad_segs=1)  # Roughly 10 meters around the point
+    except GEOSException:
+        raise HTTPException(status_code=400, detail={"coords": f"Invalid or unparseable wkt provided: {coords}"})
+    except TypeError:
+        raise HTTPException(status_code=400, detail={"coords": f"Invalid geometric type: {point.geom_type}"})
```
lukas-phaf, line 112: Why not just raise the `HTTPException` directly on line 107?

api/routers/edr.py
lukas-phaf: Thanks. Can we get all the `Query` ones? There is a bunch more, lines:
35, 69, 72, etc...

api/routers/edr.py
```diff
@@ -89,13 +97,24 @@ async def get_data_location_id(
 )
 async def get_data_position(
     coords: str = Query(..., example="POINT(5.179705 52.0988218)"),
-    parameter_name: str = Query(..., alias="parameter-name", example="dd,ff,rh,pp,tn"),
-    datetime: str | None = None,
+    parameter_name: Annotated[str | None, Query(alias="parameter-name", example="dd,ff,rh,pp,tn")] = None,
+    datetime: Annotated[str | None, Query(alias="datetime", example="2022-12-31T00:00Z/2023-01-01T00:00Z")] = None,
     f: str = Query(default="covjson", alias="f", description="Specify return format."),
```
lukas-phaf, line 102: You only need to give `alias` if the HTTP name is different from the Python name. Because `parameter-name` is not a valid Python name...

api/routers/edr.py
```diff
@@ -73,7 +77,11 @@ async def get_data_location_id(
     get_obs_request = dstore.GetObsRequest(
         filter=dict(
             platform=dstore.Strings(values=[location_id]),
-            instrument=dstore.Strings(values=list(map(str.strip, parameter_name.split(",")))),
+            **(
+                {"instrument": dstore.Strings(values=list(map(str.strip, parameter_name.split(","))))}
+                if parameter_name
+                else {}
+            ),
```
lukas-phaf, line 84: How about something like this (untested):
```
    filter = dict(platform=dstore.Strings(values=[location_id]))
    if parameter_name:
        dict["instrument"] = dstore.Strings(values=list(map(str.strip, parameter_name.split(","))))
    get_obs_request = dstore.GetObsRequest(
        filter=filter,
        temporal_interval=dstore.TimeInterval(start=range[0], end=range[1]) if range else None,
    )
```

api/dependencies.py
```diff
@@ -2,6 +2,7 @@
 from datetime import timedelta
```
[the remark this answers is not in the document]
lukas-phaf: Ok.

api/routers/edr.py
fjugipe: Has the same effect and I guess that would be easier to read & add new filters. I guess I'll do:
```
    filter = {}
    if parameter_name: 
        filter=dict(instrument=dstore.Strings(values=list(map(str.strip, parameter_name.split(",")))))

    range = get_datetime_range(datetime)
    get_obs_request = dstore.GetObsRequest(
        filter=filter,
        spatial_area=dstore.Polygon(
            points=[dstore.Point(lat=coord[1], lon=coord[0]) for coord in poly.exterior.coords]
        ),
        temporal_interval=dstore.TimeInterval(start=range[0], end=range[1]) if range else None,
    )
```
with the area endpoint then.

api/routers/edr.py
fjugipe: The general `Exception` would catch it and then raise a new `HTTPException` with incorrect details. The general `Exception` was left there as a safety measure because even though shapely uses `GEOSExpection` as a base class, `ValueError`  can be raised in some cases.

api/routers/edr.py
fjugipe: Sorry, I misunderstood the first comment. Added Annotated to each `Query` and `Path` as recommended by FastAPI

api/routers/edr.py
fjugipe: https://github.com/EURODEO/e-soh/pull/30/commits/f6787791d2654b40dbe37798c10f0fd85358e49c

api/routers/edr.py
lukas-phaf: Hmm, I don't like communication by Exception over a couple of lines, but I see the point now. Another option would be to just reraise the HTTP exception in the final handler, but not sure that is much nicer...

api/routers/edr.py
```diff
@@ -106,16 +123,30 @@ async def get_data_position(
...
+        raise HTTPException(status_code=400, detail={"coords": f"Invalid geometric type: {poly.geom_type}"})
+    except Exception:
+        raise HTTPException(
+            status_code=400, detail={"coords": f"Unexpected error occurred during wkt parsing: {coords}"}
+        )
+
     range = get_datetime_range(datetime)
+    filter = {}
+    if parameter_name:
+        filter = dict(instrument=dstore.Strings(values=list(map(str.strip, parameter_name.split(",")))))
```
lukas-phaf, line 147: For consistency?
```suggestion
        filter["instrument"] = dstore.Strings(values=list(map(str.strip, parameter_name.split(","))))
```

lukas-phaf (approved):

PR #32, opened 2024-02-09
Added parameter_name and related fields

jo-asplin-met-no:
The parameter_name field is intended to be a human-readable, representation of the "minimum" set of metadata fields that in practice will often suffice for deciding what is measured by a given time_series.

It will have the following form:

  &lt;standard name&gt;\_&lt;level&gt;\_&lt;function&gt;\_&lt;period&gt;

Note that currently only standard_name is an official ACDD attribute.

Note also that two different time series may have the same parameter_name (and differ e.g. in the instrument field).


protobuf/datastore.proto
```diff
@@ -99,6 +99,10 @@ message TSMetadata {
   string unit = 21;
   string instrument = 22;
   string instrument_vocabulary = 23 [json_name = "instrument_vocabulary"];
+  string level = 24;
```
lukas-phaf, line 102: I at first thought that you forgot to add the code for these fields...

protobuf/datastore.proto
```diff
@@ -99,6 +99,10 @@ message TSMetadata {
...
+  string period = 25;
+  string function = 26;
+  string parameter_name = 27;
```
lukas-phaf, line 105: Why do we want to ingestor to provide this? It should be constructed from the other four, won't it just violate the single source of truth principle?

To me this is just an EDR API level construct, which can/should be constructed on the fly.

lukas-phaf, line 105: Why do we want to ingestor to provide `parameter_name`? It can be constructed from the other four, won't storing it just violate the single source of truth principle?

To me this is just an EDR API level construct, which can/should be constructed on the fly.

lukas-phaf: Hmm, but how to effectively query on it, if it is not an explicit field in the datastore.

datastore/migrate/data/migrations/1707380318_add_columns.down.sql
```diff
@@ -0,0 +1,11 @@
+ALTER TABLE time_series
+	DROP COLUMN parameter_name,
+	DROP COLUMN function,
+	DROP COLUMN period,
+	DROP COLUMN level,
+	DROP CONSTRAINT unique_main,
```
lukas-phaf, line 6: I changed my mind, it is probably a good idea to have guards in the "down" script, in case of a partial run of "up" script.

datastore/migrate/data/migrations/1707380318_add_columns.down.sql
```diff
@@ -0,0 +1,11 @@
...
+	ADD CONSTRAINT unique_main UNIQUE NULLS NOT DISTINCT (version, type, title, summary, keywords,
+		keywords_vocabulary, license, conventions, naming_authority, creator_type, creator_name,
+		creator_email, creator_url, institution, project, source, platform, platform_vocabulary,
+		standard_name, unit, instrument, instrument_vocabulary, link_href, link_rel, link_type,
+		link_hreflang, link_title);
```
lukas-phaf, line 11: What does it mean in practice to drop and add a constraint on an existing table. I guess it means rebuilding an index. Not a big deal now, as we are not yet in production, but if you have a big running database, it might be very slow.

protobuf/datastore.proto
jo-asplin-met-no: You have a point there. How about the parameter_name being optional from the ingestor (but all four components mandatory)? The datastore will then if necessary construct the parameter_name on the fly. The premise is of course that the ingestor and datastore always agree on the format of the parameter_name (component order, separator char etc.).

johtoblan: The ingestion constructs the parameter_name for the datastore / EDR api, it will not be provided by the dataproducers. The four components will be provided

datastore/migrate/data/migrations/1707380318_add_columns.down.sql
jo-asplin-met-no: Do we have a choice, though? There is a way to modify the constraint as such, but not by adding or removing individual columns it seems:

```
ALTER CONSTRAINT constraint_name [ DEFERRABLE | NOT DEFERRABLE ] [ INITIALLY DEFERRED | INITIALLY IMMEDIATE ]
```

Exactly how often will this migration thingy run, though?

protobuf/datastore.proto
Teddy-1000: After some conversations with other colleagues at MET I also think we should limit the encouragement of parsing the parameter-name by the end user. And treat this more like a key then a dynamic field. Also forcing end users to use the collection information to gather available parameter-names. Making the service more robust for changes to the rules related to the parameter-name field.

lukas-phaf: For me, the `parameter-name` is just an identifier which makes the EDR API more usable. So ideally it is only exposed in queries to the datastore. So I think it shouldn't be constructed by the ingestion, but somehow constructed by the datastore.

So it shouldn't be a parameter of `PutObservations`, but it should be a filter on `GetObservations`. Either calculated on the fly as part of `GetObserations`, or added to the table as part of `PutObservations`. @jo-asplin-met-no , what do you think is best?

datastore/migrate/data/migrations/1707380318_add_columns.down.sql
lukas-phaf: It's fine for now, but if we make further changes to this while we are running in production, we should test for the downtime.

protobuf/datastore.proto
lukas-phaf: My preference would be if `parameter_name` is **not** part of PutObservations. See also my comment above.

lukas-phaf: If we want to make the service robust to changes to the rules related to the `parameter-name` field, the ingestor should not be aware of it. Ideally it wouldn't even be in the database tables, but only part of the filter in `GetObservations`. But I don't know if that is feasible.

jo-asplin-met-no: So how about this: 1) The datastore keeps both parameter_name and the four components - in the proto file and in the database (I don't think the redundancy in itself would be something to worry about unless we keep changing the parameter_name format very often, which I guess is unlikely?). 2) The datastore constructs the parameter_name itself from the four components, i.e. the ingestor is only required to provide the four components (i.e. any parameter name provided by the ingestor will be overwritten by the datastore anyway!). 3) From the perspective of the EDR and Records APIs the parameter_name and the components can be used as equivalent ways of searching for requested time series. In particular, the datastore will (in GetObservations and GetTSAttrGroups) not treat parameter_name in any special way.

Teddy-1000: After a short discussion between me and @lukas-phaf, we suggest to go with the simplest solution for now. Leaving the generation of parameter-name to the ingest. Treating the parameter-name field simply as a key, and this also leaves gettings the available parameter-names and search for parameter-names a very simple action for the datastore. We can run in to a situation where we get different parameter-names for the same observation, but this would only happen if we choose to update how we generate the parameter-name.
If the rest here then agree, the PR can be left as is and this conversation can be resolved.

Teddy-1000 (approved):
LGTM

PR #35, opened 2024-02-13
Add parameter information to /collections and /locations endpoint

lukas-phaf:
Also improved the data endpoints to return the same parameter info as the /collections and /locations endpoints.

TODO:

- [x] Fix integration tests
- [x] The code is currently hardcoded to get a specific range of data, which corresponds to 1 row in the KNMI dataset. This should be removed and replaced by the "get last value" functionality that Jo is working on (https://github.com/EURODEO/e-soh/pull/41)


api/routers/edr.py
```diff
@@ -42,47 +49,41 @@ async def get_locations(
...
+            properties={"parameter-name": sorted(platform_parameters[station_id])},
             geometry=Point(
                 type="Point",
-                coordinates=(ts.obs_mdata[0].geo_point.lon, ts.obs_mdata[0].geo_point.lat),
+                coordinates=platform_coordinates[station_id],
             ),
-        )  # HACK: Assume loc the same
-        for ts in sorted(ts_response.observations, key=lambda ts: ts.ts_mdata.platform)
+        )
+        for station_id in sorted(platform_parameters.keys())  # Sort by station_id
```
jo-asplin-met-no, line 106: The code looks fine, but just a general comment: It is decided yet what can go into a station_id? Could it be e.g. a WIGIOS ID (which I assume would be globally unique, no?), but also e.g. a national station ID (which then would be only be unique in combination with e.g. a standard country code)?

jo-asplin-met-no: And what about sensors that are moving (and then by definition not stationary at all) - should they still be identified by a station_id? (Maybe deal with that later anyway I guess)

api/routers/edr.py
```diff
@@ -42,47 +49,41 @@ async def get_locations(
...
-    for obs in sorted(ts_response.observations, key=lambda obs: obs.ts_mdata.platform):
-        platform_standard_names[obs.ts_mdata.platform].add(obs.ts_mdata.instrument)
+    platform_parameters: DefaultDict[str, Set[str]] = defaultdict(set)
+    platform_coordinates: Dict[str, Tuple[float, float]] = {}
+    all_parameters: Dict[str, Parameter] = {}
+    for obs in ts_response.observations:
+        parameter = Parameter(
+            description={"en": obs.ts_mdata.title},
+            observedProperty=ObservedProperty(
+                id=f"https://vocab.nerc.ac.uk/standard_name/{obs.ts_mdata.standard_name}",
```
jo-asplin-met-no, line 63: Nice for FAIRness!

api/routers/edr.py
```diff
@@ -42,47 +49,41 @@ async def get_locations(
...
+                label={"en": obs.ts_mdata.instrument},
+            ),
+            unit=Unit(label={"en": obs.ts_mdata.unit}),
+        )
+        platform_parameters[obs.ts_mdata.platform].add(obs.ts_mdata.instrument)
+        # HACK
+        platform_coordinates[obs.ts_mdata.platform] = (obs.obs_mdata[0].geo_point.lon, obs.obs_mdata[0].geo_point.lat)
```
jo-asplin-met-no: Why is this a hack? Because it doesn't support the case where obs location may change within a time series?

api/routers/edr.py
lukas-phaf: I was thinking about combining `naming_authority` with `platform`, where `naming_authority` is `WIGOS` (or empty) if it is a WIGOS ID?

lukas-phaf: On moving stations, they probably should not show up in the `/locations` endpoints at all, but not clear yet how to filter them. I guess we can deal with that later.

api/routers/edr.py
lukas-phaf: I rewrote the code a little bit, in the new version I collect all the coordinates for a station, and check if each station has a unique location. There is a problem if different timeseries on one platform have different coordinates.

jo-asplin-met-no: Right, so the premise here is that this particular kind of stations will be just that: stationary? And then we need to somehow deal with mobile sensors in some other way altogether, i.e. in different part of the code?

api/metadata_endpoints.py
```diff
@@ -29,11 +39,44 @@ def get_landing_page(request):
     )
 
 
-def get_collection_metadata(request) -> Collection:
+async def get_collection_metadata(request, is_self) -> Collection:
+    # TODO: Try to remove/lower duplication with /locations endpoint
+    start_datetime = Timestamp()
+    start_datetime.FromDatetime(datetime(2022, 12, 31, 23, 50))  # HACK: Force only one point for test data
```
lukas-phaf, line 45: Remove this hack!!

datastore/data-loader/client_knmi_station.py
```diff
@@ -19,6 +19,7 @@
 from google.protobuf.timestamp_pb2 import Timestamp
 from parameters import knmi_parameter_names
 
+regex_level = re.compile(r"first|second|third|grass|[0-9]+(\.[0-9]+)?m|(Level )[0-9]+|(?<=\()[ A-z]+(?=\))", re.IGNORECASE)
```
Jeffrey-Vervoort-KNMI, line 22: Not all of these levels are valid. But they make the parameters uniquely identifiable. So that parameter definitions have a distinguishable ID. But this still results in issues for e.g. /locations when the same parameter has a different description. Although this is less likely to occur because this only happens when a parameter for a different station has a different description as it only takes the latest obs metadata description.

api/metadata_endpoints.py
```diff
@@ -30,11 +37,40 @@ def get_landing_page(request):
     )
 
 
-def get_collection_metadata(request) -> Collection:
+async def get_collection_metadata(request, is_self) -> Collection:
+    # TODO: Try to remove/lower duplication with /locations endpoint
```
Jeffrey-Vervoort-KNMI, line 41: Due to the `platform_parameters` and `platform_coordinates` in `\locations`, I think that this is challenging and can negatively impact the readability of the code.

api/routers/edr.py
```diff
@@ -27,40 +34,74 @@
...
-    response_model=FeatureCollection,
+    response_model=EDRFeatureCollection,
     response_model_exclude_none=True,
 )
 # We can currently only query data, even if we only need metadata like for this endpoint
 # Maybe it would be better to only query a limited set of data instead of everything (meaning 24 hours)
 async def get_locations(
     bbox: Annotated[str, Query(example="5.0,52.0,6.0,52.1")]
-) -> FeatureCollection:  # Hack to use string
+) -> EDRFeatureCollection:  # Hack to use string
```
Teddy-1000, line 44: Why do we need a custom FeatureCollection? Is regular geojson not sufficient?

Teddy-1000 (approved):
Left a comment regarding the EDRFeatureCollection, otherwise the PR looks good.

api/routers/edr.py
Jeffrey-Vervoort-KNMI: The focus of regular GeoJSON is representing spatial data. EDR GeoJSON adds custom properties that help users discover what is available in the request. See here for the [optional requirements](https://docs.ogc.org/is/19-086r4/19-086r4.html#toc9).

The GeoJSON package implements the GeoJSON standards. The models inside the package ignore extra properties. Therefore it cannot be used to add the optional properties of [EDR Feature Collection GeoJSON](https://schemas.opengis.net/ogcapi/edr/1.1/openapi/schemas/edr-geojson/edrFeatureCollectionGeoJSON.yaml). In addition, I think it is nice to have validation on the parameters which a custom model makes possible.

I do believe this is not the right place for this model, I think a better location would be https://github.com/KNMI/edr-pydantic. I will make a PR on that repository to add the model.

Jeffrey-Vervoort-KNMI: Yes, I think this can be used for other endpoints too.

lukas-phaf: I am not sure the `/items` parameter needs the `parameters` field in the FeatureCollection, as the Features returned are timeseries, which basically are parameters. So it would be duplicate in some sense.

Teddy-1000: Aha, so this is a parameter field for a FeatureCollection, which is unnecesarry for `/items` since that `FeatureCollection` only contains `Features`. Sorry I am a bit slow here.

lukas-phaf: Let's see what makes the most sense for the different use of GeoJSON in the E-SOH API. :-)
@Jeffrey-Vervoort-KNMI, I agree that it makes sense to have this in the edr-pydantic package, but maybe best if we move that once we have the E-SOH side working, so we don't have to make multiple iterations?

Jeffrey-Vervoort-KNMI: That is fine for me, then we merge this one with the custom class, and update the edr-pydantic package later on.

datastore/integration-test/response/collection/area/200/data_within_an_area_with_two_parameters.json
```diff
@@ -52,35 +52,43 @@
                 ]
             },
             "parameters": {
-                "relative_humidity_2.0_mean_PT1M": {
+                "wind_speed_10_mean_PT10M": {
```
lukas-phaf, line 55: What is the order that is used? As this is not alphabetical. Is it fixed? Collection endpoints does seem to be ordered? Should really be the same.

api/formatters/covjson.py
```diff
@@ -85,7 +98,9 @@ def _collect_data(ts_mdata, obs_mdata):
         (o.obstime_instant.ToDatetime(tzinfo=timezone.utc), float(o.value)) for o in obs_mdata
     )  # HACK: str -> float
     (times, values) = zip(*tuples)
-    param_id = ts_mdata.parameter_name
-    unit = ts_mdata.unit
+    # param_id = ts_mdata.parameter_name
+    # standard_name = ts_mdata.standard_name
+    # title = ts_mdata.title
+    # unit = ts_mdata.unit
```
lukas-phaf, line 104: These lines can go, right?

.github/workflows/ci.yml
```diff
@@ -54,6 +54,9 @@ jobs:
       - name: Start docker stack
         run: DYNAMICTIME=false LOTIME=1000-01-01T00:00:00Z HITIME=9999-12-31T23:59:59Z docker compose up -d --wait --wait-timeout 120
 
+      - name: Test client runs without errors
+        run: DYNAMICTIME=false LOTIME=1000-01-01T00:00:00Z HITIME=9999-12-31T23:59:59Z docker compose run --rm client
+
```
lukas-phaf, line 59: This is now run twice (see a couple of lines down). Any idea why?

Jeffrey-Vervoort-KNMI: As the client is part of the profile test in the docker compose. I ran into the problem that the integration tests failed because locally they had the client, but in the pipeline not. 
I wanted to move it upwards in the pipeline. But instead of moving, I copied it.

api/metadata_endpoints.py
```diff
@@ -30,11 +37,40 @@ def get_landing_page(request):
...
+    ts_request = dstore.GetObsRequest(temporal_mode="latest")
+    ts_response = await get_obs_request(ts_request)
+
+    # Sadly, this is a different parameter as in the /locations endpoint, due to an error in the EDR spec
+    # See: https://github.com/opengeospatial/ogcapi-environmental-data-retrieval/issues/427
+    all_parameters: Dict[str, Parameter] = {}
+
+    for obs in ts_response.observations:
+        parameter = Parameter(
+            description=obs.ts_mdata.title,
```
Teddy-1000, line 51: Sorry I did not catch this earlies. But I do not think `ts_mdata.title` can be used here, as it will most likely be different for every time series. What we really need is the standard_name description from nerc vocab.

Teddy-1000: We also run in to the same issue in `/locations`

.github/workflows/ci.yml
lukas-phaf: Originally client did not leave any data... but now it does, and the integration test depend on this data? Do we want that?

api/metadata_endpoints.py
lukas-phaf: So you are hitting the "mutiple definitions" 500 error?

Shall we just use `standard_name` for description for now?
I am guessing that `obs.ts_mdata.instrument` for label (3 lines down) is also not correct.

.github/workflows/ci.yml
Jeffrey-Vervoort-KNMI: I had my doubts about that too, the reason that I added it is that it shows current issues such as the empty string as default for gRPC instead of NULL. But I can change it as well to not include it.

api/metadata_endpoints.py
lukas-phaf: To get back to this, I don't think the description from the nerc vocab would do, as that is a description of the standard_name, not of this specific parameter.

All the different air_temperatures now have the same description, which doesn't make to much sense. Some measure it at grass height, some at 2.0m. And there can have different average times. The would all be described as "Air temperature is the bulk temperature of the air, not the surface (skin) temperature.".

What we need in the metadata is a field that describes the parameter as measured, independent of the station.

lukas-phaf: BTW, why are all the `ts_mdata.title` different in your case. Do you include the station name in them?

PR #43, opened 2024-02-19
Added json endpoint.

shamlymajeed:
Request body in json format is verified using pydantic models


ingest/api/datastore.py
```diff
@@ -0,0 +1,143 @@
...
+def putObsRequest(put_obs_request):
+    try:
+        channel = grpc.aio.insecure_channel(
+            f"{os.getenv('DATASTORE_HOST', 'localhost')}:{os.getenv('DATASTORE_PORT', '50050')}"
+        )
+        grpc_stub = dstore_grpc.DatastoreStub(channel)
+    except grpc._channel._InactiveRpcError as e:
+        logger.error("Failed to connect:", e)
+    else:
+        logger.info("Connection established successfully.")
```
Teddy-1000, line 137: This should log what connection was successful.

ingest/api/ingest.py
Teddy-1000, line 1: This file seems to have docstrings for the old layout of ingest, and should be updated.

ingest/api/ingest.py
```diff
@@ -31,57 +32,67 @@ def __init__(
...
-                    self.dstore.ingest(msg)
-                    self.mqtt.send_message(msg)
-                except grpc.RpcError:
-                    self.dstore.is_channel_ready()
-                    pass
-                except Exception:
-                    pass
+                    ingest(msg)
+                except grpc.RpcError as v_error:
+                    return "Failed to ingest" + "\n" + str(v_error), 500
```
Teddy-1000, line 87: the grpc error might contain internal information and probably should not always be returned to the user. There is no need to inform of how data is passed to the datastore, but we should tell the user about what they could do to resolve the error. I.e "API could not reach datastore" is one option, and the user should know they did nothing wrong.

ingest/api/ingest.py
```diff
@@ -31,57 +32,67 @@ def __init__(
...
+                except Exception as e:
+                    return "Failed to ingest" + "\n" + str(e), 502
```
Teddy-1000, line 89: We should never return a raw error the the user. If we have an unexpected error, the user should just get a 500 error with the message "Internal server error". This should be done with the fastAPI HTTPException.

ingest/api/ingest.py
```diff
@@ -99,7 +110,7 @@ def _decide_input_type(self, message) -> str:
                 logger.critical(f"Unknown filetype provided. Got {message.split('.')[-1]}")
                 raise ValueError(f"Unknown filetype provided. Got {message.split('.')[-1]}")
```
Teddy-1000, line 111: Errors in the API should raised with HTTPException

ingest/api/main.py
Teddy-1000, line 1: If the API should raise an error, do not return a response with error codes, use the HTTPException.

ingest/api/main.py
```diff
@@ -0,0 +1,70 @@
...
+        return Response(status_message=response, status_code=status)
+
+    except Exception as e:
+        # No specfic exceptions are thrown from ingest
+        # So catch all and send a generic response back
+        return Response(status_message=str(e), status_code=500)
+
+
+@app.post("/json")
+async def post_json(request: JsonMessageSchema, input_type: str):
```
Teddy-1000, line 63: input_type need an explanation and example values. Is input_type needed here?

ingest/api/messages.py
```diff
@@ -29,14 +33,14 @@ def build_message(file: [object], input_type: str, uuid_prefix: str, schema_path
         json_msg["id"] = message_uuid
         json_msg["properties"]["metadata_id"] = message_uuid
         json_msg["properties"]["data_id"] = message_uuid
-
         json_msg["properties"]["pubtime"] = datetime.now(timezone.utc).isoformat()
         try:
```
Teddy-1000, line 37: Validation should no longer be needed with jsonschema, but handled by pydantic.

shamlymajeed: We are making the metadata_id, data_id after receiving the actual data .so don't we need to verify whether these fields exist or not?

ingest/api/model.py
```diff
@@ -0,0 +1,316 @@
...
+
+
+class JsonMessageSchema(BaseModel):
+    type: Type
+    geometry: Union[Geometry, Geometry1]
+    properties: Properties
+    links: List[Link] = Field(..., min_items=1)
+    version: str
+
+    def dict(self, *args, **kwargs) -> Dict[str, Any]:
```
Teddy-1000, line 306: Why do you need this method?
If you need the dict from a model use the model_dump method inherited from BaseModel

shamlymajeed: for nested objects to a dictionary, we need to customize the serialization process

Teddy-1000 (approved):

Issue #37, opened 2024-02-14
Regard only a subset of the TSMetadata fields as the time series key


jo-asplin-met-no:
It should be possible that only some of the fields in TSMetadata define a unique time series. This allows "non-key" fields of a given time series to be updated without having to create a completely new time series. Typical use cases include fixing typos and updating the contents of the _links_ field (list of alternative landing pages to the underlying time series etc.).

jo-asplin-met-no:
Candidates for such a "unique combination" subset:

- naming_authority
- platform
- standard_name
- level
- function
- period
- instrument

All other fields may then be updated (i.e. **overwritten**!) by PutObservations calls made by the ingestor.

jo-asplin-met-no:
Just a reality check: What's the rationale of keeping the metadata_id with each observation? I understand that it works like a reference to the time series to which the observation belongs. But wouldn't that be useful only in the case where it would be practical (or even required) to handle individual observations in isolation / out-of-context? And would that really be the case here? I mean, isn't it always the case that the time series in question - with all the ts-specific metadata - is readily available in all systems that process data from E-SOH? And why do we then need this reference explicitly in each obs? Couldn't we simply remove it? (may not an entirely transferable example, but it's hard to come up with a use case for keeping a ref to the image explicitly in every single pixel of a digital image)

Teddy-1000:
For the metadata feature endpoint we need an ID of some kind to reference a single feature, in our case a time series. We are not shipping all metadata with the covjson data, so we need a simple way of looking up the time series. We do not need to explicitly keep the metadata_id with all observations though.

jo-asplin-met-no:
Then I suggest to move metadata_id to TSMetadata since this is metadata that doesn't change between observations within the same time series. I.e. something like this:

> diff --git a/protobuf/datastore.proto b/protobuf/datastore.proto
> index a78771e..14d360b 100644
> --- a/protobuf/datastore.proto
> +++ b/protobuf/datastore.proto
> @@ -103,6 +103,7 @@ message TSMetadata {
>    string period = 25;
>    string function = 26;
>    string parameter_name = 27 [json_name = "parameter_name"];
> +  string metadata_id = 28 [json_name = "metadata_id"];
>    // --- END string metadata -----------------
>  }
> 
> @@ -123,7 +124,6 @@ message ObsMetadata {
>    string id = 6;
>    string data_id = 7 [json_name = "data_id"];
>    string history = 8;
[...]

PR #46, opened 2024-02-27
Issue 37 tsmdata subset key

jo-asplin-met-no:
Issue: https://github.com/EURODEO/e-soh/issues/37


datastore/datastore/storagebackend/postgresql/postgresql.go
```diff
@@ -21,6 +22,82 @@ func (sbe *PostgreSQL) Description() string {
...
+	query := `
+		SELECT pg_get_constraintdef(c.oid)
+		FROM pg_constraint c
+		JOIN pg_namespace n ON n.oid = c.connamespace
+		WHERE conrelid::regclass::text = 'time_series'
+			AND conname = 'unique_main'
+			AND contype = 'u'
+	`
+
+	row := sbe.Db.QueryRow(query)
```
lukas-phaf, line 40: Can you give an example of the query output?

datastore/datastore/storagebackend/postgresql/postgresql.go
```diff
@@ -79,6 +164,38 @@ func getTSMdataCols() []string {
...
+
+	for _, col := range getTSMdataCols() { // start with all columns
+		colSet[col] = struct{}{}
+	}
+
+	for _, col := range getTSMdataColsUnique() { // remove columns of the unique_main constraint
+		delete(colSet, col)
+	}
+
+	// return remaining columns as a string
```
lukas-phaf, line 187:
```suggestion
	// return remaining columns as a string slice
```

datastore/datastore/storagebackend/postgresql/postgresql.go
```diff
@@ -21,6 +22,82 @@ func (sbe *PostgreSQL) Description() string {
...
+	cols := getTSMdataCols()
+
+	formats := make([]string, len(cols))
+	for i := 0; i < len(cols); i++ {
+		formats[i] = "$%d"
+	}
+
+	updateExpr := []string{}
+	for _, col := range getTSMdataColsUniqueCompl() {
+		updateExpr = append(updateExpr, fmt.Sprintf("%s = EXCLUDED.%s", col, col))
```
lukas-phaf, line 76: We don't care about the order of the different columns here, correct? If I understand Go correctly, the return value of `etTSMdataColsUniqueCompl` will be randomized, correct?

datastore/datastore/storagebackend/postgresql/putobservations.go
```diff
@@ -75,14 +82,42 @@ func getTSColVals(tsMdata *datastore.TSMetadata) ([]interface{}, error) {
...
+// save unnecessary database access.
+//
 // Returns (ID, nil) upon success, otherwise (..., error).
-func getTimeSeriesID(
+func upsertTS(
 	db *sql.DB, tsMdata *datastore.TSMetadata, cache map[string]int64) (int64, error) {
 
-	colVals, err := getTSColVals(tsMdata)
+	colVals2 := map[string]interface{}{}
+	colVals, err := getTSColVals(tsMdata, colVals2) // column values for U+UC
```
lukas-phaf, line 120: Asymmetry in how `colVals` and `colVals2` are returned. Return both as return values?

Or, why not return `colVals` and `colsValsUnique` directly from `getTsColVals()`, both as slice? So that you don't have to call `getTSColValsUnique` later?

datastore/migrate/data/migrations/1708586491_reduce_unique_constraint.up.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE time_series
+	DROP CONSTRAINT IF EXISTS unique_main,
+	ADD CONSTRAINT unique_main UNIQUE NULLS NOT DISTINCT (naming_authority, platform,
+		standard_name, instrument, level, period, function);
```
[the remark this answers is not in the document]
jo-asplin-met-no: You mean for now always just replace the single pair of up/down files whenever we make a change? I wouldn't mind. Similarly, we should be allowed to [modify field numbers](https://protobuf.dev/programming-guides/proto3/) freely before we enter real production. What do you think?

datastore/datastore/storagebackend/postgresql/putobservations.go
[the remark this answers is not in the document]
jo-asplin-met-no: Yes, the documentation of what the cache is needs to be updated according to your suggestion. Change coming up.

datastore/datastore/storagebackend/postgresql/postgresql.go
jo-asplin-met-no:
```
$ PGPASSWORD=mysecretpassword psql -h localhost -p 5433 -U postgres -d data -t -c "SELECT pg_get_constraintdef(c.oid) FROM pg_constraint c JOIN pg_namespace n ON n.oid = c.connamespace WHERE conrelid::regclass::text = 'time_series' AND conname = 'unique_main' AND contype = 'u'"
 UNIQUE NULLS NOT DISTINCT (naming_authority, platform, standard_name, instrument, level, period, function)
```

datastore/datastore/storagebackend/postgresql/postgresql.go
jo-asplin-met-no: Correct! And correct!

jo-asplin-met-no: (I think it's a good thing that golang randomizes map iteration, since it forces developers to treat a map as an unordered structure where you can't rely on some arbitrary system-dependent order)

datastore/migrate/data/migrations/1708586491_reduce_unique_constraint.up.sql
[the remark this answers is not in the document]
lukas-phaf: I would be on favor of both. Go into production without the historical baggage. @Teddy-1000, what do you think?
Also, is this a problem for the deployments we are running?

datastore/datastore/storagebackend/postgresql/putobservations.go
jo-asplin-met-no: I agree with the first comment, so will return both as return values. Change coming up.

But I don't see the value of refactoring to derive colVals and colValsUnique from the same function, neither in terms of performance or code readability. (But if they were derived from the same function to begin with, I probably wouldn't have bothered to split into two functions).

datastore/migrate/data/migrations/1708586491_reduce_unique_constraint.up.sql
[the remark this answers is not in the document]
Teddy-1000: This is no problem for the test deployments, we would lose the data currently stored, but I do not see that as an issue and is expected in our current stage of development. I think it's nice to "start fresh" when we get to production.

datastore/datastore/storagebackend/postgresql/postgresql.go
lukas-phaf: I think it would help to have this in an example, as it helps to understand the code.

datastore/datastore/storagebackend/postgresql/putobservations.go
lukas-phaf: The reason for the second point is that `getTSColVals` function now returns the column values in two formats, a slice and a dictionary. This feels wrong to me. As the dict is only used to make the colValsUnique, I suggested to return that directly, restoring symmetry. Another option is to let `getTSColCals` return only the dict, and derive the two slices from that.

lukas-phaf: But, it's not a big deal. :-)

datastore/datastore/storagebackend/postgresql/postgresql.go
jo-asplin-met-no: Good point - will do.

datastore/datastore/storagebackend/postgresql/putobservations.go
jo-asplin-met-no: Well I see your point more clearly now. And I agree. I'll see if I can improve it.

jo-asplin-met-no: I like your suggestion of letting getTSColCals return only the dict, and derive the two slices from that. So that's what I'll implement.

jo-asplin-met-no: But then I ran into an issue probably caused by golang randomizing map iteration. So will have to sort ...

jo-asplin-met-no: I'm afraid I have to revert to the original implementation of having getTSColVals return two structures: one map and one array. The problem is that other parts of the code assumes a predefined order of the ts fields (in particular that the non-string fields (not handleable with reflection) need to be processed explicitly in a certain order (so far the links field is the only one in this category)). Returning just a map and then deriving the two slices from that messes up the order. It's probably possible to fix this somehow, bit I suspect it will take a while to get it right everywhere, and I propose to do this in a later refactoring. So being a bit pragmatic for now. I will add a comment about it in the code, though.

jo-asplin-met-no: Hm ... looking again at your earlier suggestion:

> Or, why not return colVals and colsValsUnique directly from getTsColVals(), both as slice? So that you don't have to call getTSColValsUnique later?

I'll try that. Seems to make good sense.

jo-asplin-met-no: (Sorry for once again getting your point quite late in the conversation :) )

Jeffrey-Vervoort-KNMI (approved):

datastore/datastore/storagebackend/postgresql/putobservations.go
```diff
@@ -63,32 +70,68 @@ func getTSColVals(tsMdata *datastore.TSMetadata) ([]interface{}, error) {
...
+// upsertTS retrieves the ID of the row in table time_series that matches tsMdata wrt.
+// the fields - U - defined by constraint unique_main, inserting a new row if necessary.
+//
+// If the row already existed, the function ensures that the row is updated with the tsMdata
+// fields - UC - that are not in U (i.e. the complement of U).
+//
+// The ID is first looked up in a cache (where the key consists of all fields (U + UC)) in order to
+// save unnecessary database access. In other words, a cache hit means that the row for
+// time series not only existed, but was also already fully updated according to tsMdata.
+// And vice versa: a cache hit means the row either didn't exist at all or wasn't fully updated
```
lukas-phaf, line 123: miss?

jo-asplin-met-no: Definitely!

jo-asplin-met-no: Will piggyback this fix on the next PR. 🐖

PR #52, opened 2024-03-11
Use GetTSAGRequest() for /collections endpoint.

lukas-phaf:
For the full 200K timeseries in my test for /collections, gRPC message size goes down from 44 MB to 4289 bytes. Runtime from 2250 ms to 88 ms.

### DIFF
--- a/api/metadata_endpoints.py
+++ b/api/metadata_endpoints.py
@@ -1,3 +1,4 @@
+import logging
 from typing import Dict
 
 import datastore_pb2 as dstore
@@ -17,7 +18,11 @@ from edr_pydantic.parameter import Parameter
 from edr_pydantic.unit import Unit
 from edr_pydantic.variables import Variables
 from fastapi import HTTPException
-from grpc_getter import get_obs_request
+from grpc_getter import getTSAGRequest
+
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
 
 
 def get_landing_page(request):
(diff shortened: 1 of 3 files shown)


api/metadata_endpoints.py
```diff
@@ -17,7 +18,11 @@
 from edr_pydantic.unit import Unit
 from edr_pydantic.variables import Variables
 from fastapi import HTTPException
-from grpc_getter import get_obs_request
+from grpc_getter import getTSAGRequest
+
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
```
Teddy-1000, line 25: Great to get logging in, but should we not grab the same log-level as set when running the server?

lukas-phaf: Maybe I messed things up. Logging wasn't working at all for me, so I just added this stuff to get it working, and thought it might be useful, so I left it in.

What is the right way to do this? Where do I get it from the server? What server is that, uvicorn?

Teddy-1000: It seems that the loglevel set when starting uvicorn is set globally. Gunicorn was not as easy. I think we should open a separate issue for implementing logging.

Teddy-1000 (approved):

api/metadata_endpoints.py
lukas-phaf: So leave it like it is, or strip it out?

Teddy-1000: I think you can just leave it for now.

Issue #42, opened 2024-02-16
Define endpoint and capabilites for metadata discovery and serach


Teddy-1000:
Summary originally written by @GijsbertKruithof . 
We had a meeting today 15-02-2024 with [@Phaf, Lukas (KNMI)](mailto:lukas.phaf@knmi.nl), [@amundi@met.no](mailto:amundi@met.no), [@Jo Asplin](mailto:joa@met.no) and [@jukka.pelli@fmi.fi](mailto:jukka.pelli@fmi.fi) about how to provide metadata via the API.

Especially if we do this with an OGC Records API or with the OGC EDR API.

## Outcome

- We can use the EDR Items endpoint for time series metadata
- The discovery metadata of the datasets will be made available via a JSON file. The message from the MQTT broker will point to an url where the JSON file can be fetched.
- For station metadata the user will be pointed to a OSCAR entry, which will be based on the WIGOS identifier – if present in the metadata.
 

### Main reasons for using EDR:

The EDR Items endpoint has exactly the same features structure in both API’s. [EDR: /collections/{collectionId}/items](https://docs.ogc.org/is/19-086r6/19-086r6.html#_81254e3a-d734-4195-ad15-ce64e208728c),  [OGC Records: /collections/{collectionid}/items](https://ogcapi.ogc.org/records/overview.html)
But using the OGC EDR is easier for the user to understand, and easier for us to make and maintain

 

## To be done:

- The time series metadata needs to be added to the EDR API to be searchable.
- The WIGOS identifier needs to be translated into an URL in the API output
- The discovery metadata on the Dataset needs to be made available in a JSON file.

Teddy-1000:
I will start implementing the search for time series in the EDR API.
Any thoughts on where we should put the dataset Discovery Metadata file. I don't think it can be under /items?

Teddy-1000:
After some looking in to features vs Records items search, the main difference seems to be the support CQL querying.
In a features endpoint, we need to add every parameter we want to be searchable. In records we would rather implement a CQL2  expression for filtering.
Both are of course possible to implement, any thoughts on which we should go with? @lukas-phaf @jo-asplin-met-no @johtoblan

vegark73:
Could this be a first atempt for the text required?

"title": "Aditional and subhourly observations from European area, including synoptic observations distrinuted over GTS",
```

```
[...]

PR #56, opened 2024-03-14
Issue 42 discovery metadata

Teddy-1000:
Closing #42 
The E-SOH dataset is added under `/collections/observations/dataset` and is undocumented in OpenAPI.
Time series is also searchable in `/collections/observations/items` with most of the metadata keywords being searchable. No wildcard support yet, but all keywords can be added as a comma separated list.


datastore/integration-test/response/metadata/200/items_within_area.json
```diff
@@ -0,0 +1,1643 @@
+{
```
lukas-phaf, line 1: I think this file is to big to maintain. Can we use a smaller bbox?

api/formatters/covjson.py
```diff
@@ -27,7 +27,9 @@
 
 def make_parameter(ts_mdata):
     return Parameter(
-        description={"en": ts_mdata.title},
+        description={
+            "en": "For detailed description see observedProperty id"
+        },  # ts_mdata.title}, # title is unique for each TS, and does not describe a parameter
```
lukas-phaf, line 32: The description parameter is not mandatory. I think it is better to just drop it then have this hardcoded string.

But just referring to the observedProperty id is not good enough, as that describes for example air temperature, where there should be someplace to give more detailed info.

At KNMI we have a parameter level description (e.g. "Ambient Temperature 1.5m 10 Min Minimum") which is useful to have here for the user. Could we somehow have metadata which we say has to be at parameter level? Or maybe it could be constructed from other fields.

Can you give an example of how this comes out unique for each TS at Met Norway?

api/formatters/geojson.py
```diff
@@ -0,0 +1,63 @@
+from fastapi import HTTPException
+from geojson_pydantic import Feature
+from geojson_pydantic import FeatureCollection
+from geojson_pydantic import Point
+
+
+def _make_properties(ts):
+    return {
```
lukas-phaf, line 8: Leave out empty/not set fields?

datastore/integration-test/response/metadata/200/items_get_single_platform.json
```diff
@@ -0,0 +1,1643 @@
+{
```
lukas-phaf, line 1: Can we pick a platform with less timeseries? `06225` has only three.

To be honest, I was a bit shocked by that this PR has way more lines then the whole API up to now... :-(

api/formatters/geojson.py
```diff
@@ -0,0 +1,63 @@
+from fastapi import HTTPException
+from geojson_pydantic import Feature
+from geojson_pydantic import FeatureCollection
+from geojson_pydantic import Point
+
+
+def _make_properties(ts):
```
lukas-phaf, line 7: Do we want to use reflection to avoid having to change this every time the fields change? Or leave it for now in the interest of progress?

api/metadata_endpoints.py
```diff
@@ -48,7 +48,7 @@ async def get_collection_metadata(request, is_self) -> Collection:
 
     for obs in ts_response.observations:
         parameter = Parameter(
-            description=obs.ts_mdata.title,
+            description="For detailed description see observedProperty id",
```
lukas-phaf, line 51: Same comment as for the `description` in the covjson.

api/utilities.py
```diff
@@ -52,21 +52,18 @@ def get_datetime_range(datetime_string: str | None) -> Tuple[Timestamp, Timestam
     return start_datetime, end_datetime
 
 
+@cached(ttl=600)
 async def get_current_parameter_names(ttl_hash=None):
```
lukas-phaf, line 56: Get rid of the hash, as we now control TTL directly. The current setup will just fill up memory with no positive effect...

api/utilities.py
```diff
@@ -52,21 +52,18 @@ def get_datetime_range(datetime_string: str | None) -> Tuple[Timestamp, Timestam
...
+    unique_parameter_names = dstore.GetTSAGRequest(attrs=["parameter_name"])
+    unique_parameter_names = await getTSAGRequest(unique_parameter_names)
 
-        return set([i.combo.standard_name for i in unique_parameter_names.groups])
-
-    return await async_helper(ttl_hash)
+    return set([i.combo.parameter_name for i in unique_parameter_names.groups])
 
 
 async def verify_parameter_names(parameter_names: list) -> None:
```
lukas-phaf, line 69: Do we want to use this, what is the advantage over just passing it into the store?

And if we do, do we want to add caching? How long does the GetTSAGRequest take?
I always try to avoid using caching. In this cache, the cached verify_parameter_names might actually be out of alignment with the parameters available in the call into the store (as that is not cached).

api/routers/edr.py
```diff
@@ -40,14 +39,19 @@
 # We can currently only query data, even if we only need metadata like for this endpoint
 # Maybe it would be better to only query a limited set of data instead of everything (meaning 24 hours)
 async def get_locations(
-    bbox: Annotated[str, Query(example="5.0,52.0,6.0,52.1")]
+    bbox: Annotated[str | None, Query(example="5.0,52.0,6.0,52.1")] = None
```
lukas-phaf, line 42: I would prefer to leave this mandatory. The specs are unclear on what it should be.
Forcing people to provide a `bbox` might get the to be have. Otherwise it is to easy to just not provide the parameter, which in most cases will put unnecessary load on the system.

api/routers/edr.py
```diff
@@ -233,9 +237,10 @@ async def get_data_area(
         )
 
     range = get_datetime_range(datetime)
-    # await verify_parameter_names(parameter_name)
+    parameter_name = parameter_name.split(",") if parameter_name else None
+    await verify_parameter_names(parameter_name)
```
lukas-phaf, line 241: This will conflict with PR from Jeffrey.

api/routers/feature.py
```diff
@@ -0,0 +1,205 @@
...
+from utilities import get_datetime_range
+
+router = APIRouter(prefix="/collections/observations")
+
+
+@router.get(
+    "/items", tags=["Collection items"], response_model=Feature | FeatureCollection, response_model_exclude_none=True
+)
+async def search_timeseries(
+    bbox: Annotated[str | None, Query(example="5.0,52.0,6.0,52.1")] = None,
```
lukas-phaf, line 22: Let's make this mandatory, even though that probably contradicts the spec. I think this will lead to better behavior of our users...

api/routers/feature.py
```diff
@@ -0,0 +1,205 @@
...
+    ] = None,
+    period: Annotated[
+        str | None, Query(description="Duration of collection period in ISO8601", example="PT10M")
+    ] = None,
+    function: Annotated[
+        str | None, Query(description="Aggregation function used to sample obersved property", example="maximum")
+    ] = None,
+    f: Annotated[
+        formatters.Metadata_Formats, Query(description="Specify return format")
+    ] = formatters.Metadata_Formats.geojson,
```
lukas-phaf, line 60: Does this follow the OGC API Feature standard for search, or are we going for this for simplicity?
If yes, could you point me to the relevant section?

api/routers/feature.py
```diff
@@ -0,0 +1,205 @@
...
+    instrument: Annotated[str | None, Query(description="Instrument Id")] = None,
+    level: Annotated[
+        str | None,
+        Query(description="Instruments height above ground or distance below surface, in meters", example=2),
+    ] = None,
+    period: Annotated[
+        str | None, Query(description="Duration of collection period in ISO8601", example="PT10M")
+    ] = None,
+    function: Annotated[
+        str | None, Query(description="Aggregation function used to sample obersved property", example="maximum")
```
lukas-phaf, line 56:
```suggestion
        str | None, Query(description="Aggregation function used to sample observed property", example="maximum")
```

api/routers/feature.py
```diff
@@ -0,0 +1,205 @@
...
+):
+    if bbox:
+        left, bottom, right, top = map(str.strip, bbox.split(","))
+        poly = geometry.Polygon([(left, bottom), (right, bottom), (right, top), (left, top)])
+    if datetime:
+        range = get_datetime_range(datetime)
+
+    obs_request = dstore.GetObsRequest(
+        filter=dict(
+            metadata_id=dstore.Strings(values=ids.split(",") if ids else ids),
```
lukas-phaf, line 70: `else None`?

api/routers/feature.py
Teddy-1000: Section [7.15.5]([https://docs.ogc.org/is/17-069r4/17-069r4.html#_parameters_for_filtering_on_feature_properties) allows for defining parameters for filtering on feature properties. But my understanding of this section might be wrong.

api/routers/edr.py
Teddy-1000: It is my understanding that bbox functionality is not yet implemented in geoweb, and geoweb will need to be able to get all locations for now. I will investigate it, and make it mandatory if geoweb can provide the bbox.

api/routers/feature.py
Teddy-1000: For the most part I do agree, but in the case where on is looking up on or more specific platform(s) it would be user friendly to not have to provide the bbox.
Does fastAPI have "OneOf" functionality we could use? Or would we have to create a model for the request?

api/formatters/geojson.py
Teddy-1000: I have rewritten it.

api/utilities.py
Teddy-1000: If gRPC would report that the parameter name given had no matches, this becomes unnecessary.
I would like to differentiate for the user between parameter names with errors, and if the chosen query just had no matches.

api/formatters/geojson.py
Teddy-1000: Implemented dropping of empty fields.

api/utilities.py
lukas-phaf: Ok, fair enough, we can keep it. What about the second paragraph of my comment? Is the potential inconsistency worth the performance improvement?

Call to get the valid parameters takes 2 ms.

api/utilities.py
```diff
@@ -52,21 +53,18 @@ def get_datetime_range(datetime_string: str | None) -> Tuple[Timestamp, Timestam
...
-
-        return set([i.combo.standard_name for i in unique_parameter_names.groups])
+    unique_parameter_names = dstore.GetTSAGRequest(attrs=["parameter_name"])
+    unique_parameter_names = await getTSAGRequest(unique_parameter_names)
 
-    return await async_helper(ttl_hash)
+    return set([i.combo.parameter_name for i in unique_parameter_names.groups])
 
 
 async def verify_parameter_names(parameter_names: list) -> None:
```
lukas-phaf, line 70: This function assumes a list of parameters is coming in, while actually a comma separated list is passed.

lukas-phaf, line 70: Make sure to `await` everywhere this is called.

api/routers/feature.py
lukas-phaf: Agreed!

api/utilities.py
Teddy-1000: I will update the function and remove the cache.

api/routers/feature.py
```diff
@@ -0,0 +1,140 @@
...
+    bbox: Annotated[str | None, Query(example="5.0,52.0,6.0,52.1")] = None,
+    datetime: Annotated[
+        str | None,
+        Query(
+            example="2022-12-31T00:00Z/2023-01-01T00:00Z",
+            description="E-SOH database only contains data from the last 24 hours",
+        ),
+    ] = None,
+    ids: Annotated[str | None, Query(description="Comma separated list of time series ids")] = None,
+    parameter_name: Annotated[str | None, Query(alias="parameter-name", description="E-SOH parameter name")] = None,
```
lukas-phaf, line 38: All of the parameters support comma seperated lists. Do we need to mention this for each one?

api/routers/feature.py
```diff
@@ -0,0 +1,140 @@
...
+        raise HTTPException(400, detail="Have to set at least one of bbox or platform.")
+    if bbox:
+        left, bottom, right, top = map(str.strip, bbox.split(","))
+        poly = geometry.Polygon([(left, bottom), (right, bottom), (right, top), (left, top)])
+    if datetime:
+        range = get_datetime_range(datetime)
+
+    obs_request = dstore.GetObsRequest(
+        filter=dict(
+            metadata_id=dstore.Strings(values=ids.split(",") if ids else None),
```
lukas-phaf, line 79: I think we should do the split + strip trick so we allow spacing around the comma's.

api/routers/feature.py
```diff
@@ -0,0 +1,140 @@
...
+    return formatters.metadata_formatters[f](time_series)
+
+
+@router.get("/dataset", tags=["E-SOH dataset"], include_in_schema=False)
+async def get_dataset_metadata():
+    # need to get spatial extent.
+    spatial_request = dstore.GetExtentsRequest()
+    extent = await get_spatial_extent(spatial_request)
+    dynamic_fields = {
+        "spatial_extents": [
```
lukas-phaf, line 124: I think a GeoJSON polygon has to be closed:
https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.6

I think there also needs to be an extra layer of square brackets:
https://datatracker.ietf.org/doc/html/rfc7946#appendix-A.3

api/routers/feature.py
```diff
@@ -0,0 +1,140 @@
...
+            [extent.spatial_extent.left, extent.spatial_extent.bottom],
+            [extent.spatial_extent.right, extent.spatial_extent.bottom],
+            [extent.spatial_extent.right, extent.spatial_extent.top],
+            [extent.spatial_extent.left, extent.spatial_extent.top],
+        ],
+        "temporal_extents": [
```
lukas-phaf, line 130: I guess there is no standard for time ranges in GeoJSON

api/formatters/covjson.py
```diff
@@ -27,7 +27,9 @@
 
 def make_parameter(ts_mdata):
     return Parameter(
-        description={"en": ts_mdata.title},
+        description={
+            "en": f"{ts_mdata.standard_name} at {ts_mdata.level}m {ts_mdata.period} {ts_mdata.function}",
+        },  # ts_mdata.title}, # title is unique for each TS, and does not describe a parameter
```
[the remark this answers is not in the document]
Teddy-1000: Removed

api/routers/feature.py
Teddy-1000: No, will remove.

api/routers/feature.py
Teddy-1000: Implemented split_and_strip function, and updated where it was needed.

api/formatters/geojson.py
```diff
@@ -0,0 +1,43 @@
...
```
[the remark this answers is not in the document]
lukas-phaf: This will be solved by the PR from Jo.

api/routers/edr.py
```diff
@@ -255,10 +261,9 @@ async def get_data_area(
...
 
     range = get_datetime_range(datetime)
-    # await verify_parameter_names(parameter_name)
     if parameter_name:
-        parameter_name = parameter_name.split(",")
-        parameter_name = list(map(lambda x: x.strip(), parameter_name))
+        parameter_name = split_and_strip(parameter_name)
+        await verify_parameter_names(parameter_name)
     request = dstore.GetObsRequest(
         filter=dict(parameter_name=dstore.Strings(values=parameter_name if parameter_name else None)),
```
lukas-phaf, line 268: Should set `parameter_name` to None in the dict if no parameters are given, not but None in the `Strings` like we do now.

api/routers/edr.py
```diff
@@ -151,9 +158,8 @@ async def get_data_location_id(
...
     range = get_datetime_range(datetime)
     if parameter_name:
-        parameter_name = parameter_name.split(",")
-        parameter_name = list(map(lambda x: x.strip(), parameter_name))
-    # parameter_name = verify_parameter_names(parameter_name) # should the api verify that the parameter name is valid?
+        parameter_name = split_and_strip(parameter_name)
+        await verify_parameter_names(parameter_name)
     request = dstore.GetObsRequest(
         filter=dict(
             parameter_name=dstore.Strings(values=parameter_name),
```
lukas-phaf, line 165: Should set `parameter_name` to None in the dict if no parameters are given.
We now pass in an empty string, which works (as that matches anything), but it is not really correct.

api/routers/edr.py
lukas-phaf: Will make separate issue for this.

api/routers/edr.py
lukas-phaf: Will make separate issue for this.

datastore/integration-test/response/items_single_id.json
```diff
@@ -0,0 +1,33 @@
+{
```
lukas-phaf, line 1: Can we use the id that is in `items_area_with_one_parameter_name.json` so we can drop this file?

datastore/integration-test/response/items_within_area_single_platform.json
```diff
@@ -0,0 +1,104 @@
+{
```
lukas-phaf, line 1: Can we use the same result file as for `items_get_single_platform.json` by tweaking the bbox?

datastore/integration-test/test_api.py
```diff
@@ -150,14 +150,14 @@ def test_from_a_single_collection_get_a_single_location_which_does_not_exist():
 
     expected_json = load_json("response/404_not_found.json")
 
-    assert actual_response.status_code == 404
+    assert actual_response.status_code == 400
```
lukas-phaf, line 153: Why did this change?
The file is also still named with 404.

api/test/test_ts_geojson.py
```diff
@@ -0,0 +1,30 @@
...
+        expected_json = json.load(file)
+
+    diff = DeepDiff(expected_json, actual_response.json(), **kwargs)
+    assert diff == {}
+
+
+def test_geojson_conversion():
+    ts_response = dstore.GetObsResponse()
+    with open("test/test_data/test_ts_geojson_proto_object.json") as file:
+        json_format.ParseDict(json.load(file), ts_response)
```
lukas-phaf, line 21: Can we use `Parse()` like in `create_mock_obs_response`. I think it has slightly different meaning.

api/test/test_ts_geojson.py
```diff
@@ -0,0 +1,30 @@
+import json
+
+import datastore_pb2 as dstore
+from deepdiff import DeepDiff
+from formatters.geojson import convert_to_geojson
+from google.protobuf import json_format
+
+
+def actual_response_is_expected_response(actual_response, expected_path, **kwargs):
+    # file_path = Path(Path(__file__).parent, expected_path).resolve()
```
lukas-phaf, line 10: Drop this line?

api/test/test_ts_geojson.py
```diff
@@ -0,0 +1,30 @@
...
+    assert diff == {}
+
+
+def test_geojson_conversion():
+    ts_response = dstore.GetObsResponse()
+    with open("test/test_data/test_ts_geojson_proto_object.json") as file:
+        json_format.ParseDict(json.load(file), ts_response)
+
+    result = convert_to_geojson(ts_response)
+    print(result)
```
lukas-phaf, line 24: Drop this?

api/test/test_ts_geojson.py
```diff
@@ -0,0 +1,30 @@
+import json
+
+import datastore_pb2 as dstore
+from deepdiff import DeepDiff
+from formatters.geojson import convert_to_geojson
+from google.protobuf import json_format
+
+
+def actual_response_is_expected_response(actual_response, expected_path, **kwargs):
```
lukas-phaf, line 9: I don't like the asymmetry of one coming in as a file name, and the other as a Python dict...

api/test/test_data/test_ts_expected_geojson.json
```diff
@@ -0,0 +1,122 @@
...
+                "type": "Point",
+                "coordinates": [
+                    5.8723225499118,
+                    52.0548617826
+                ]
+            },
+            "properties": {
+                "version": "",
+                "type": "",
+                "summary": "",
```
lukas-phaf, line 16: Empty strings should no longer be here, correct? So why doesn't this test fail?

datastore/integration-test/test_api.py
Teddy-1000: It changed due to verify_parameter_name, I updated the filename.

lukas-phaf (approved):

PR #66, opened 2024-03-19
Issue 65 - Make the bbox for locations optional.

Jeffrey-Vervoort-KNMI:
- Makes the bbox optional
- Does requires the gRPC size to be increased

### DIFF
--- a/api/routers/edr.py
+++ b/api/routers/edr.py
@@ -53,9 +53,9 @@ response_fields_needed_for_data_api = [
 # We can currently only query data, even if we only need metadata like for this endpoint
 # Maybe it would be better to only query a limited set of data instead of everything (meaning 24 hours)
 async def get_locations(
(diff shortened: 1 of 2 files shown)


api/routers/edr.py
```diff
@@ -53,15 +53,18 @@
...
+        poly = geometry.Polygon([(left, bottom), (right, bottom), (right, top), (left, top)])
+        spatial_area = dstore.Polygon(
+            points=[dstore.Point(lat=coord[1], lon=coord[0]) for coord in poly.exterior.coords],
+        )
 
     ts_request = dstore.GetObsRequest(
-        spatial_area=dstore.Polygon(
-            points=[dstore.Point(lat=coord[1], lon=coord[0]) for coord in poly.exterior.coords],
-        ),
+        spatial_area=spatial_area,
```
lukas-phaf, line 67: Re-order this so we never add `spatial_area` as `None`.

lukas-phaf (approved):

PR #74, opened 2024-03-20
Issue 60 radius query

jo-asplin-met-no:


api/routers/edr.py
```diff
@@ -286,7 +286,8 @@ async def get_data_area(
         )
 
     request = dstore.GetObsRequest(
-        spatial_area=dstore.Polygon(
+        filter=dict(parameter_name=dstore.Strings(values=parameter_name if parameter_name else None)),
```
lukas-phaf, line 289: I think something went wrong with the rebase here. `parameter_name` is now added twice, see line 298.

jo-asplin-met-no: Right! I was a bit unsure of that line. Will remove it ...

protobuf/datastore.proto
```diff
@@ -51,6 +51,11 @@ message Polygon { // horizontal area; three or more points
   repeated Point points = 1;
 }
 
+message Circle { // horizontal circle area
+  Point center = 1;
+  double radius = 2;
```
lukas-phaf, line 56: Looking at the code this is in km, but this is not specified. I would have expected meters if not specified, but better to make it explicit.

lukas-phaf (commented):
Looks good!

This will allow us to implement the EDR [radius](https://docs.ogc.org/is/19-086r6/19-086r6.html#_fa3fccfc-df4b-43c0-9195-243d3594226e) query. Without that, this code will not really get exercised. Should we add a test to client.py to do that, until the EDR `radius` endpoint is implemented?

On the `radius` endpoint, looking at the spec, this will also require to implement the supported `within_units` in the collections radius `data_queries` section, see https://docs.ogc.org/is/19-086r6/19-086r6.html#_1b54f97a-e1dc-4920-b8b4-e4981554138d

Would you mind making the issue for that?

jo-asplin-met-no:
> Looks good!
> 
> This will allow us to implement the EDR [radius](https://docs.ogc.org/is/19-086r6/19-086r6.html#_fa3fccfc-df4b-43c0-9195-243d3594226e) query. Without that, this code will not really get exercised. Should we add a test to client.py to do that, until the EDR `radius` endpoint is implemented?
> 
> On the `radius` endpoint, looking at the spec, this will also require to implement the supported `within_units` in the collections radius `data_queries` section, see https://docs.ogc.org/is/19-086r6/19-086r6.html#_1b54f97a-e1dc-4920-b8b4-e4981554138d
> 
> Would you mind making the issue for that?

Good points. Issue coming up.

protobuf/datastore.proto
jo-asplin-met-no: Do you prefer to have the unit in the variable name (`radius` -> `radius_km`), or does it suffice to add a comment like in this case:
```
message Point { // horizontal position
  double lat = 1; // latitude degrees in range [-90, 90]
  double lon = 2; // longitude degrees in range [-180, 180]
}
```
Either way I guess we should ensure consistency so that we follow the same practice everywhere?

jo-asplin-met-no: I guess with Issue #109 the unit won't be in the name at all since it is defined separately in `radius_unit`. So only documenting in the comment for now in order to minimize modifications required soon by implementation of #109.

lukas-phaf:
Sorry, I wasn't very clear. My question was to make an issue for the API to add the `/radius` query type. I think we are ok with initially supporting only `km`.

jo-asplin-met-no:
> Sorry, I wasn't very clear. My question was to make an issue for the API to add the `/radius` query type. I think we are ok with initially supporting only `km`.

Updated Issue #109.

lukas-phaf (approved):

PR #67, opened 2024-03-19
Issue 59 - Added platform_name and synchronized definition order

jo-asplin-met-no:

Issue #48, opened 2024-03-04
Determine what metadata should be included in covjson.


Teddy-1000:
Current there are very little metadata included with the covjson. This should be kept to a minimum set, but probably need to be expanded some.
I would like some feedback on the set of metadata that should be included with the covjson.

Teddy-1000:
Currently there are very not any discovery metadata included. I think at least a title and metadata_id should be included.

johtoblan:
> This should be kept to a minimum set
Why? Is not the covjson hierarchical?

PR #75, opened 2024-03-20
Issue 68 - Get platform name for /locations endpoint from datastore

### LINKED ISSUE #68: Platform name from datastore to /locations

fjugipe:
Closes #68 

- Get platform name from datastore
- Update details url to wmo json endpoint
- Update test data loader and test data

Requires PR #67 before merge


Teddy-1000:
I think we should also update the covjson.py and geojson.py to put the platform in the name field if platform_name is not set.

fjugipe:
Changed to use platform_name instead of name. We are already using platform for id, and platform_vocabulary for Oscar linkage, so I think platform_name would be consistent. Thoughts on this?

api/routers/edr.py
```diff
@@ -80,9 +81,13 @@ async def get_locations(
     ts_response = await get_obs_request(ts_request)
 
     platform_parameters: DefaultDict[str, Set[str]] = defaultdict(set)
+    platform_names: Dict[str, str] = defaultdict(str)
```
lukas-phaf, line 84: To check for a feature having more then one platform names (they might be inconsistent in different time series), store them in a `Dict[str, Set[str]]`, and do the same check as for the coordinates?

datastore/integration-test/response/items_get_single_platform.json
```diff
@@ -24,9 +24,9 @@
                 "institution": "Royal Netherlands Meteorological Institute (KNMI)",
                 "project": "",
```
lukas-phaf, line 25: Why doesn't this test break? As the output is now without the empty strings, correct?

fjugipe: I didn't touch the tests, only updated the test data. Seems like this is duplicate / older version of `items_within_area_single_platform.json` and is not used on any tests.

api/routers/edr.py
fjugipe: Refactored in https://github.com/EURODEO/e-soh/pull/75/commits/0ef8818d457f7d494cefec908fee2cf05ed4b600

api/formatters/geojson.py
```diff
@@ -8,13 +8,16 @@ def _make_properties(ts):
...
     )
-    # This should also be compatible with future when name is added to datastore
-    if "name" not in ts_metadata:
-        ts_metadata["name"] = ts.ts_mdata.platform  # TODO: grab proper name when implemented in proto
+
+    if "platform_name" in ts_metadata:
+        ts_metadata["name"] = ts_metadata["platform_name"]
+        del ts_metadata["platform_name"]
+    else:
+        ts_metadata["name"] = f'platform-{ts_metadata["platform"]}'
```
lukas-phaf, line 20: Keep this `platform_name`, as we are listing time series here. Using `name` would be confusing.

fjugipe: Changed back to `platform_name` in `/items`

api/formatters/covjson.py
```diff
@@ -101,4 +101,7 @@ def _collect_data(ts_mdata, obs_mdata):
     )  # HACK: str -> float
     (times, values) = zip(*tuples)
 
+    if not ts_mdata.platform_name:
+        ts_mdata.platform_name = f"platform-{ts_mdata.platform}"
```
lukas-phaf, line 105: I can't find where this is actually used? Does it end up in an output? None of the data json's are changed...

If it should be in the output, it should also be added to `response_fields_needed_for_data_api`.

lukas-phaf (commented):
Sorry, one more comment. Besides that, it looks fine.

api/formatters/covjson.py
fjugipe: It was made based on @Teddy-1000 https://github.com/EURODEO/e-soh/pull/75#issuecomment-2012075933 and I thought it would be related to https://github.com/EURODEO/e-soh/issues/48.

It makes sure that the `ts_mdata` has a platform_name when collecting data needed to convert to `CoverageJSON`. But at the moment the converter does not use that part of `ts_mdata`  for a response, so it does not end up in any output. That is why I left it out from `response_fields_needed_for_data_api` in purpose for now to keep the message size at bay.

lukas-phaf: I would prefer to leave it out then. To much risk of not updating `response_fields_needed_for_data_api` when this gets added to the converter, and always ending up with the default...

fjugipe: Removed in https://github.com/EURODEO/e-soh/pull/75/commits/aae3ab5fff2c4c26087a48d4175eed431ab17e20

PR #82, opened 2024-03-21
Issue 73 - Clean up request filter building

### LINKED ISSUE #73: [API] Clean up how we build the GetObsRequest filter when there are optional parameters
In `edr.py`, correctly handle the case of `parameter_name=None`, see lines 165 and 268. Probably best to build the filter dict, only adding something if it is provided. Do the same for other optional parameters.

fjugipe:
Closes #73

- Build filter & datetime when needed for obs request
- Update API unit tests and notify when tests fail
- Update pre-commit hooks config

Note /locations request filtering is fixed in #66

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -143,7 +143,7 @@ jobs:
       - name: Run Tests
         run: |
           cd api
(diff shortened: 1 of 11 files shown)


fjugipe:
Check #82 instead.

PR #88, opened 2024-03-22
Test can break

lukas-phaf:
Broken API unit test do not actually break the CI actions. Fix the tests (half of them) that were actually broken...


fjugipe:
I did the same fixes in #82 already, but implemented the pipefail with different implementation.

lukas-phaf:
> I did the same fixes in #82 already, but implemented the pipefail with different implementation.

Hmmm, great teamwork. ;-)

I think the pipefail approach I have is better, as it is global across all tasks...

With respect to the actual tests, I don't know which is better.
I did base all the tests of the KNMI test data now, and put the actual queries used in the code.

fjugipe:
> > I did the same fixes in #82 already, but implemented the pipefail with different implementation.
> 
> Hmmm, great teamwork. ;-)
> 
> I think the pipefail approach I have is better, as it is global across all tasks...
> 
> With respect to the actual tests, I don't know which is better. I did base all the tests of the KNMI test data now, and put the actual queries used in the code.

I agree on the pipefail 👍 

And to the actual tests: These are more up to date and for example have mocks to functions which didn't exist in the main branch when I made mine. So I think we should go with these.

PR #92, opened 2024-03-22
Add conformance endpoint.

Jeffrey-Vervoort-KNMI:


Teddy-1000 (approved):

api/metadata_endpoints.py
```diff
@@ -36,12 +36,29 @@ def get_landing_page(request):
...
+            Link(href=f"{request.url}conformance", rel="conformance", title="Conformance Declaration in JSON"),
             Link(href=f"{request.url}collections", rel="data", title="Collections metadata in JSON"),
         ],
     )
 
 
+def get_conformance() -> ConformanceModel:
+    return ConformanceModel(
+        conformsTo=[
+            "https://www.opengis.net/spec/ogcapi-common-1/1.0/conf/core",  # A2 - required
```
lukas-phaf, line 48: In the OGC EDR example, the links are slightly different:
https://docs.ogc.org/is/19-072/19-072.html#_d5e2189c-9836-4af5-9d91-69b02cd1eab1

The first one for example is:
"http://www.opengis.net/spec/ogcapi-common-1/1.0/req/core"
Do we now what the differences is, and why did we pick the ones with `conf`?

Also, what about the others in that example, do we conform to those (like the `landing-page` one)?

Also, there seems to be a `json` and `html` in both `common` and `edr`. Why did we pick one and not the other?

api/metadata_endpoints.py
```diff
@@ -36,12 +36,29 @@ def get_landing_page(request):
...
+            "https://www.opengis.net/spec/ogcapi-edr-1/1.1/req/collections",  # A3 - required
+            "https://www.opengis.net/spec/ogcapi-edr-1/1.1/req/queries",  # A4
+            "https://www.opengis.net/spec/ogcapi-edr-1/1.1/conf/core",  # B2
+            "https://www.opengis.net/spec/ogcapi-edr-1/1.1/conf/collections",  # B3
+            "https://www.opengis.net/spec/ogcapi-edr-1/1.1/conf/json",  # B4
+            "https://www.opengis.net/spec/ogcapi-edr-1/1.1/conf/geojson",  # B5
+            "https://www.opengis.net/spec/ogcapi-edr-1/1.1/conf/covjson",  # B7
+            "https://www.opengis.net/spec/ogcapi-edr-1/1.1/conf/html",  # B8
+            "https://www.opengis.net/spec/ogcapi-edr-1/1.1/conf/oas30",  # B9
```
lukas-phaf, line 57: Maybe my question is really: Why did we list these, and did we check we actually conform?

api/metadata_endpoints.py
Jeffrey-Vervoort-KNMI: They fixed the URLs. The example initially contained broken links and URLs with conf. I tested the URLs and now all the URLs in the example of the spec work. Will look into what the difference between `conf` and `req` is.

Jeffrey-Vervoort-KNMI: At the moment the examples for 1.1 are back at `conf` again: https://docs.ogc.org/is/19-086r6/19-086r6.html#_19183664-f3c9-4df4-b6e8-4afa9c54ac08

Jeffrey-Vervoort-KNMI: Requirements belong to a Conformance Class. Each Requirements Class defines a functional requirement. Conformance classes are advertised through the `/conformance` path.

lukas-phaf (approved):

PR #77, opened 2024-03-20
Issue 71 - Get Extent from the datastore.

Jeffrey-Vervoort-KNMI:
- Gets the spatial and temporal extent from the datastore.

PR #93, opened 2024-03-25
Issue 73 - Refactor building ObsRequest optional filter parameters

### LINKED ISSUE #73: [API] Clean up how we build the GetObsRequest filter when there are optional parameters

fjugipe:
Closes #73 

- Build filter & datetime only when needed for obs request

Note /locations request filtering already fixed in #66

### DIFF
--- a/api/routers/edr.py
+++ b/api/routers/edr.py
@@ -158,18 +158,22 @@ async def get_data_location_id(
 ):
     # TODO: There is no error handling of any kind at the moment!
     #  This is just a quick and dirty demo
-    range = get_datetime_range(datetime)
+    request = dstore.GetObsRequest(
+        filter=dict(
+            platform=dstore.Strings(values=[location_id]),
+        ),
+        included_response_fields=response_fields_needed_for_data_api,
+    )
+
     if parameter_name:
         parameter_name = split_and_strip(parameter_name)
         await verify_parameter_names(parameter_name)
-    request = dstore.GetObsRequest(
-        filter=dict(
-            parameter_name=dstore.Strings(values=parameter_name),
-            platform=dstore.Strings(values=[location_id]),
-        ),
-        temporal_interval=(dstore.TimeInterval(start=range[0], end=range[1]) if range else None),
-        included_response_fields=response_fields_needed_for_data_api,
-    )
+        request.filter["parameter_name"].values.extend(parameter_name)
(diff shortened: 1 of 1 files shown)


api/routers/edr.py
```diff
@@ -158,18 +158,22 @@ async def get_data_location_id(
...
     )
+
+    if parameter_name:
+        parameter_name = split_and_strip(parameter_name)
+        await verify_parameter_names(parameter_name)
+        request.filter["parameter_name"].values.extend(parameter_name)
+
+    if datetime:
+        range = get_datetime_range(datetime)
+        request.temporal_interval.CopyFrom(dstore.TimeInterval(start=range[0], end=range[1]))
```
lukas-phaf, line 175: You can't assign messages fields directly, and this is why you use this "trick", right?

Accroding to [this](https://github.com/EURODEO/e-soh/pull/77), you can also set `start` and `end` fields directly, would that be clearer?

fjugipe: You can't directly assign a new value. That is why I made a new TimeInterval object and used `CopyFrom` to assign to it at the same time. If the values or fields are already there, then they can be assigned.

I could make it like this, if you think it is more clearer? 
```suggestion
        request.temporal_interval.CopyFrom(dstore.TimeInterval())
        request.temporal_interval.start = range[0]
        request.temporal_interval.end = range[1]
```

lukas-phaf: Sorry, put the wrong link *and* was reading it wrong.

The link is:
https://protobuf.dev/reference/python/python-generated/#embedded_message

This works:
```
        request.temporal_interval.start.CopyFrom(range[0])
        request.temporal_interval.end.CopyFrom(range[1])
```

I'll leave it up to you to what you like best. But I would change the first line to tuple unpacking, so something like:
```
        start, end = get_datetime_range(datetime)
        request.temporal_interval.start.CopyFrom(start)
        request.temporal_interval.end.CopyFrom(end)
```

fjugipe: I see your point now and refactored it 👍

lukas-phaf (approved):

PR #96, opened 2024-03-25
Fix CoverageCollection parameters

lukas-phaf:
The CoverageCollection returned the parameters of the first Coverage. It now returns the union of parameters of all coverages.


Teddy-1000 (approved):
Looks good to me.

Issue #97, opened 2024-03-26
Update gunicorn launch parameters


Teddy-1000:
Need to add the possibility to set "forwarded-allow-ips' to gunicorn launch parameters to forwards scheme from proxy to server.

PR #98, opened 2024-03-26
Update api docker file with new launch parameter

Teddy-1000:
Closing #97 
Updated launch parameter to allow forward of schema from proxy.


docker-compose.yml
```diff
@@ -66,6 +66,7 @@ services:
     environment:
       - DSHOST=store
       - DSPORT=50050
+      - PROXY_IP=${PROXY_IP:-}
```
lukas-phaf, line 69: Although this works, maybe it is better to set the default to the default value for the parameter (127.0.0.1), see here:
https://docs.gunicorn.org/en/stable/settings.html#forwarded-allow-ips

lukas-phaf: Also, did you check if using env variables in the `CMD` of the dockerfile works at all?
According to this, it doesn't:
https://stackoverflow.com/a/40454758

Teddy-1000: I tried to run the server, but it worked by accident. Do we need to use the list notation for the lunch CMD? If not, I have pushed a fix, that I have verified to work.

lukas-phaf: The linting seems unhappy... How about the approach here with `"echo $VAR"`:
https://docs.docker.com/reference/dockerfile/#variable-substitution

It is probably the minimum change. Assuming it works...

lukas-phaf: Also, I would just put the default in the docker-compose file, not also in the dockerfile `CMD`.

Teddy-1000: It now work, used the method you linked.

lukas-phaf: It works, but now the gunicorn doesn't get the proper shutdown command on restart/updates. It eventually just gets killed instead. This means it is slower to restart, but more importantly, requests might be stopped halfway.

I need to look at this a bit more... it seems like such a simple change.

Teddy-1000: You can just set the `FORWARDED_ALLOW_IPS` in the env. I think this is the best way to do it. If left unset, it will just be set to default.

lukas-phaf: Oh man... that is so much better. Have we been going in circles...

Only remaining question, do we want to drop the `DEBUG` loglevel now?

Teddy-1000: Yes, ofcourse. I just added it to see if the changes took effect. Will remove it.

docker-compose.yml
```diff
@@ -66,6 +66,7 @@ services:
     environment:
       - DSHOST=store
       - DSPORT=50050
+      - FORWARDED_ALLOW_IPS=${PROXY_IP:-127.0.0.1}
```
lukas-phaf, line 69: Hmm, still one more: This should not refer to `PROXY_IP` anymore? To be consistent with README?

lukas-phaf (approved):

PR #101, opened 2024-03-27
DRAFT: Change /locations logic

lukas-phaf:
This PR is just to discuss an idea.

Change /locations to first get all unique parameters, and then the actual stations, making the connection only by parameter_name.

Advantage is that it there is less data to transfer. For the KNMI test data set:
Before: One request of 220 KB
After: One request of 4 KB followed by request of 132 KB.

Disadvantage is that we can no longer check for inconsistent parameter definitions (i.e. different parameters that have the same `parameter_name`).

I think a better way of doing this would be to do this approach in the `store` directly, which means the filtering is done in the `store`, and significantly less data would be send to the API.

### DIFF
--- a/api/routers/edr.py
+++ b/api/routers/edr.py
@@ -1,4 +1,5 @@
 # For developing:    uvicorn main:app --reload
+import logging
 from collections import defaultdict
 from typing import Annotated
 from typing import DefaultDict
@@ -20,6 +21,7 @@ from formatters.covjson import make_parameter
 from geojson_pydantic import Feature
 from geojson_pydantic import Point
 from grpc_getter import get_obs_request
+from grpc_getter import get_ts_ag_request
 from shapely import buffer
 from shapely import geometry
 from shapely import wkt
@@ -28,6 +30,11 @@ from utilities import get_datetime_range
 from utilities import split_and_strip
 from utilities import validate_bbox
 from utilities import verify_parameter_names
+
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
 
 router = APIRouter(prefix="/collections/observations")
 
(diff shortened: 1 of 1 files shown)


api/routers/edr.py
```diff
@@ -56,17 +63,27 @@
...
+    # logger.info(ts_response.ByteSize())
+
+    # Sadly, this is a different parameter as in the /locations endpoint, due to an error in the EDR spec
+    # See: https://github.com/opengeospatial/ogcapi-environmental-data-retrieval/issues/427
+    all_parameters: Dict[str, Parameter] = {}
+
+    for group in ts_response.groups:
+        ts = group.combo
+        parameter = make_parameter(ts)
+        all_parameters[ts.parameter_name] = parameter
```
Jeffrey-Vervoort-KNMI, line 79: Is it possible that there is a duplicate parameter name? If so should then an exception be raised that it already exists?

api/routers/edr.py
```diff
@@ -78,28 +95,17 @@ async def get_locations(
...
-        if obs.ts_mdata.parameter_name in all_parameters and all_parameters[obs.ts_mdata.parameter_name] != parameter:
-            raise HTTPException(
-                status_code=500,
-                detail={
-                    "parameter": f"Parameter with name {obs.ts_mdata.parameter_name} "
-                    f"has multiple definitions:\n{all_parameters[obs.ts_mdata.parameter_name]}\n{parameter}"
-                },
-            )
-        all_parameters[obs.ts_mdata.parameter_name] = parameter
+        # TODO: How to check for inconsistent parameter definitions between stations
```
Jeffrey-Vervoort-KNMI, line 108: Is this possible to do on lines 76-79? Because if a parameter_name exists that is not unique, it is likely impossible to distinguish between them for stations.

Jeffrey-Vervoort-KNMI:
I think it is an improvement that we can request more data. I am curious how it impacts the API performance. Especially now that is less likely that we hit the gRPC limit.

api/routers/edr.py
lukas-phaf: Yes, that is probably the right way/place to do it.

Teddy-1000: I don't think it is possible. `make_parameter` creates the definition from variables used for parameter_name.

jo-asplin-met-no:
FYI: Based on discussions I added [this issue](https://github.com/EURODEO/e-soh/issues/107). Feel free to comment!

Issue #95, opened 2024-03-25
Add temporal filter and parameter-name to locations endpoint


Teddy-1000:
The `/locations` endpoint should make a temporal filter available.
The `/locations` should also have parameter-name filter.

lukas-phaf:
What should this select on? Only give back stations that have data in this period? Is this useful if we only store 24 hours of data? It feels like it is very unlikely for a station to be there, but not to have data?

Teddy-1000:
> What should this select on? Only give back stations that have data in this period? Is this useful if we only store 24 hours of data? It feels like it is very unlikely for a station to be there, but not to have data?

Maybe not often, but I think we should support it.

fjugipe:
Do we want to add the temporal filter to properties section of each feature when filtering with it? E.g.:
```
{
  "type": "Feature",
  "geometry": {
    "type": "Point",
    "coordinates": [
       ...,
       ...
    ]
  },
  "properties": {
    "name": "...",
    "detail": "...",
    "datetime": "2024-03-26T00:00:00Z/2024-03-26T06:00:00Z",
    "parameter-name": [...]
  },
  "id": "..."
}
```
What about the situations when there is no temporal filter, as Lukas stated we only have 24 hours of data?

lukas-phaf:
> Maybe not often, but I think we should support it. There are some edge cases where it might be useful. For example if you would like to find all locations that have had updates the last 5 minutes. This also reduces the amount of data that need to be transmitted.

Fair enough. It should be trivial to implement anyway, just pass it to the datastore, right?

lukas-phaf:
> Do we want to add the temporal filter to properties section of each feature when filtering with it? E.g.:

I am not a fan of doing this, as it implies there is only data for the requested period...

fjugipe:
> > Do we want to add the temporal filter to properties section of each feature when filtering with it? E.g.:
> 
> I am not a fan of doing this, as it implies there is only data for the requested period...

I agree with this one and the user should already know the period because based on the request. I will leave it as is for now and we can do it later if really needed.

PR #102, opened 2024-03-27
Issue 95 - parameter-name and temporal filters to `/locations`

fjugipe:
Closes https://github.com/EURODEO/e-soh/issues/95

- Add temporal filter to `/locations`
- Add parameter-name filter to `/locations`
- Add an error message if no features are found with request to `/locations`
- Add basic unit & integration tests for new functionality
- Update test data


api/test/test_edr.py
```diff
@@ -53,22 +145,24 @@ def test_get_locations_id_without_parameter_names_query():
         mock_get_obs_request.assert_called_once()
         m_args = mock_get_obs_request.call_args[0][0]
 
-        assert "instrument" not in m_args.filter
+        assert "parameter_name" not in m_args.filter
```
lukas-phaf, line 148: Checking for something not to be there is always tricky...

api/test/test_edr.py
```diff
@@ -10,6 +10,98 @@
 client = TestClient(app)
 
 
+def test_get_locations_without_query_params():
+    with patch("routers.edr.get_obs_request") as mock_get_obs_request, patch(
+        "routers.edr.verify_parameter_names"
+    ) as mock_verify_parameter_names:
+        test_data = load_json("test/test_data/test_feature_collection_proto.json")
+        compare_data = load_json("test/test_data/test_feature_collection.json")
```
lukas-phaf, line 18: This test is really about testing that the mock gets called with the correct parameters, correct?

It is probably good to have a comments around `test_data` that is is just "random" data, not appropriate for the request. And I would also drop the `compare_data` and the `assert response.json() == compare_data` for the same reason.

That part is tested later anyway.

api/test/test_edr.py
```diff
@@ -9,6 +10,124 @@
...
+    "obstime_instant",
+    "value",
+]
+
+
+def test_get_locations_without_query_params():
+    with patch("routers.edr.get_obs_request") as mock_get_obs_request, patch(
+        "routers.edr.verify_parameter_names"
+    ) as mock_verify_parameter_names:
+        # Load arbitary test data for making a mock_obs_request
```
lukas-phaf, line 44:
```suggestion
        # Load arbitrary test data for making a mock_obs_request
```

lukas-phaf (approved):

PR #104, opened 2024-04-02
Add GitHub registry upload

Teddy-1000:
Add CI to run on merge to main branch. Will build and upload to the github registry.


.github/workflows/registry_upload.yml
```diff
@@ -0,0 +1,84 @@
+#
```
Jeffrey-Vervoort-KNMI, line 1: Would we like to run the tests in the pipeline from the published images instead of those built inside the pipeline?

.github/workflows/registry_upload.yml
```diff
@@ -0,0 +1,78 @@
+name: Create and publish a Docker image
```
Jeffrey-Vervoort-KNMI, line 1: I think we want to merge the workflow with the other workflow. So that the publishing of docker images only happens when every test passes.

Teddy-1000: The way I want it set up is that only after the PR is merged, the images are published. And all the test will have to pass before the PR can be merged.

Teddy-1000: I updated the trigger to run after CI is completed. I also added a test to only build of CI succeeded.

.github/workflows/registry_upload.yml
```diff
@@ -0,0 +1,90 @@
...
+    # Sets the permissions granted to the `GITHUB_TOKEN` for the actions in this job.
+    permissions:
+      contents: read
+      packages: write
+    strategy:
+      matrix:
+        include:
+          - context: ./datastore/datastore/
+            IMAGE_NAME: datastore
+            CPU: ["arm64", "amd64"]
```
lukas-phaf, line 36: I think these `CPU` can be dropped here, as they are only used in line 50, where the default `CPU` definition from line 15 is fine (amd64 is automatically available).

lukas-phaf (approved):
One comment added, and I wonder why not also do it for branches, as these can be useful for testing also.

Approved, so we can see if it works.

.github/workflows/registry_upload.yml
Teddy-1000: Yeah, this line is never used either

PR #108, opened 2024-04-04
Issue 105 deal with multiple locations

Teddy-1000:
Draft for handling station with slightly diverging positions in `/location`.

### DIFF
--- a/api/routers/edr.py
+++ b/api/routers/edr.py
@@ -30,6 +30,7 @@ from utilities import get_datetime_range
 from utilities import split_and_strip
 from utilities import validate_bbox
 from utilities import verify_parameter_names
+from utilities import calculate_largest_postition_deviation
 
 router = APIRouter(prefix="/collections/observations")
 
@@ -97,6 +98,7 @@ async def get_locations(
         platform_coordinates[obs.ts_mdata.platform].add(
             (obs.obs_mdata[-1].geo_point.lon, obs.obs_mdata[-1].geo_point.lat)
         )
+
         # Check for inconsistent parameter definitions between stations
         # TODO: How to handle those?
         if obs.ts_mdata.parameter_name in all_parameters and all_parameters[obs.ts_mdata.parameter_name] != parameter:
@@ -112,6 +114,16 @@ async def get_locations(
     # Check for multiple coordinates or names on one station
     for station_id in platform_parameters.keys():
         if len(platform_coordinates[station_id]) > 1:
+            if (
+                calculate_largest_postition_deviation(platform_coordinates[station_id]) < 1e-4
+            ):  # all coordinates is with 10M
+                platform_coordinates[station_id] = {
+                    sorted(
+                        [i for i in platform_coordinates[station_id]],
+                        key=lambda x: (len(str(x[0])), len(str(x[1])), x[0], x[1]),
+                    )[-1]
+                }
+                continue
             raise HTTPException(
                 status_code=500,
                 detail={
--- a/api/utilities.py
+++ b/api/utilities.py
@@ -10,7 +10,9 @@ from pydantic import AwareDatetime
 from pydantic import TypeAdapter
 
 
-def get_datetime_range(datetime_string: str | None) -> Tuple[Timestamp, Timestamp] | None:
+def get_datetime_range(
+    datetime_string: str | None,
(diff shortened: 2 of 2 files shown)


api/routers/edr.py
```diff
@@ -112,6 +114,16 @@ async def get_locations(
...
+            if (
+                calculate_largest_postition_deviation(platform_coordinates[station_id]) < 1e-4
+            ):  # all coordinates is with 10M
+                platform_coordinates[station_id] = {
+                    sorted(
+                        [i for i in platform_coordinates[station_id]],
+                        key=lambda x: (len(str(x[0])), len(str(x[1])), x[0], x[1]),
+                    )[-1]
+                }
+                continue
```
lukas-phaf, line 126: Can you rewrite the flow, the continue here feels strange.
Might also be wrong, don't we need to still check the platform_names in this case?

api/utilities.py
```diff
@@ -121,3 +123,12 @@ def validate_bbox(bbox: str) -> Tuple[float, float, float, float]:
         raise HTTPException(status_code=400, detail=errors)
 
     return left, bottom, right, top
+
+
+def calculate_largest_postition_deviation(positions: list) -> float:
+    largest_deviation = 0
+    for i in positions:
+        for j in positions:
```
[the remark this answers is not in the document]
Teddy-1000: Yes, I know it does. First I assumed the list would be short, so it would be fine. But it kinds of bugs me, so I have fixed it.

api/routers/edr.py
```diff
@@ -112,6 +114,16 @@ async def get_locations(
     # Check for multiple coordinates or names on one station
     for station_id in platform_parameters.keys():
         if len(platform_coordinates[station_id]) > 1:
+            if (
+                calculate_largest_postition_deviation(platform_coordinates[station_id]) < 1e-4
+            ):  # all coordinates is with 10M
+                platform_coordinates[station_id] = {
+                    sorted(
+                        [i for i in platform_coordinates[station_id]],
+                        key=lambda x: (len(str(x[0])), len(str(x[1])), x[0], x[1]),
```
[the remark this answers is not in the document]
Teddy-1000: This was to make sure we get the number with the most precision, we can ofcourse just sort the numbers. But as you mentioned in the issue, we can end up using coordinates with lower precision.

api/routers/edr.py
Teddy-1000: Yes, in does skip a needed test. I have updated it, so it can be reached.

api/routers/edr.py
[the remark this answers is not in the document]
lukas-phaf: Fair enough. Can you add a comments explaining the `key`?

api/routers/edr.py
```diff
@@ -112,13 +114,23 @@ async def get_locations(
...
+                    )[-1]
+                }
+            else:
+                raise HTTPException(
+                    status_code=500,
+                    detail={
+                        "coordinates": f"Station with id `{station_id} "
+                        f"has multiple coordinates: {platform_coordinates[station_id]}"
```
lukas-phaf, line 131: Write something like "multiple incompatible coordinates", so it is clear what we are actually testing.

api/utilities.py
```diff
@@ -10,7 +11,9 @@
 from pydantic import TypeAdapter
 
 
-def get_datetime_range(datetime_string: str | None) -> Tuple[Timestamp, Timestamp] | None:
+def get_datetime_range(
+    datetime_string: str | None,
+) -> Tuple[Timestamp, Timestamp] | None:
```
lukas-phaf, line 16: Change not needed with 120 line length?

Teddy-1000: Yeah, it's just a local tool getting in the way. Reverted change.

setup.cfg
```diff
@@ -2,3 +2,8 @@
 ignore = E221,E226,E228,E241,W503
 max-line-length = 120
 exclude = ingest/src/esoh/*pb2*.py
+[isort]
+profile = black
+[black]
+line-length = 120
```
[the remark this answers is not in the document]
Teddy-1000: Yes, per black documentation command-line arguments should take priority, and thus respect the ones set in the pre-commit.

PR #111, opened 2024-04-08
#110: Use store `Circle` request for `/position` endpoint.

lukas-phaf:
#110

### DIFF
--- a/api/routers/edr.py
+++ b/api/routers/edr.py
@@ -22,7 +22,6 @@ from geojson_pydantic import Point
 from grpc_getter import get_obs_request
 from response_classes import CoverageJsonResponse
 from response_classes import GeoJsonResponse
-from shapely import buffer
 from shapely import geometry
 from shapely import wkt
 from shapely.errors import GEOSException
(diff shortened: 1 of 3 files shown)


api/routers/edr.py
```diff
@@ -267,7 +244,17 @@ async def get_data_position(
             detail={"coords": f"Unexpected error occurred during wkt parsing: {coords}"},
         )
 
-    return await get_data_area(coords=poly.wkt, parameter_name=parameter_name, datetime=datetime, f=f)
+    request = dstore.GetObsRequest(
+        # 10 meters around the point
+        spatial_circle=dstore.Circle(center=dstore.Point(lat=point.y, lon=point.x), radius=0.01),
```
Teddy-1000, line 249: 0.01 is approx 1.1 km? I think radius should be 1e-4 if this radius is equivalent to decimal degrees? Ref. https://en.wikipedia.org/wiki/Decimal_degrees

lukas-phaf: Radius for this gRPC parameter is in kilometers:
https://github.com/EURODEO/e-soh/blob/main/protobuf/datastore.proto#L56

Teddy-1000 (approved):

PR #127, opened 2024-05-28
Make parameter wildcards work again.

lukas-phaf:
Also added integration test for this. And call `get_current_parameter_names()` once instead of for each requested parameter.

### DIFF
--- a/api/utilities.py
+++ b/api/utilities.py
@@ -71,9 +71,13 @@ async def verify_parameter_names(parameter_names: list) -> None:
     Raises error with unknown names if any are found.
     """
     unknown_parameter_names = []
+    current_parameter_names = await get_current_parameter_names()
 
     for i in parameter_names:
-        if i not in await get_current_parameter_names():
+        # HACK: This checking breaks the wildcard support (e.g. air_temperature:*:*:*). Skip if contains wildcards
+        if "*" in i:
+            continue
+        if i not in current_parameter_names:
             unknown_parameter_names.append(i)
 
     if unknown_parameter_names:
--- a/api/routers/edr.py
+++ b/api/routers/edr.py
@@ -60,6 +60,10 @@ async def get_locations(
         str | None,
         Query(
             alias="parameter-name",
+            description="Comma seperated list of parameter names. Each consists of four components seperated by colons."
+            " The components are standard name, level in meters, aggregation function, and period. "
+            "Each of the components can be replaced by the wildcard character `*`. "
+            "To get all the air temperatures measured at 1.5 meter, use `air_temperature:1.5:*:*`.",
             example="wind_from_direction:2.0:mean:PT10M,"
             "wind_speed:10:mean:PT10M,"
             "relative_humidity:2.0:mean:PT1M,"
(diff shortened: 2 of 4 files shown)


api/utilities.py
```diff
@@ -71,9 +71,13 @@ async def verify_parameter_names(parameter_names: list) -> None:
     Raises error with unknown names if any are found.
     """
     unknown_parameter_names = []
+    current_parameter_names = await get_current_parameter_names()
```
fjugipe, line 75: I think we really should check at least that the CF parameter name is correct, otherwise the whole verify_parameter_names is obsolete because you can just wildcard it away.

I made suggestions which check the standard_name, but also allows wildcarding for everything or to use partial filtering on the portion if you just want to wildcard for example the function / height. E.g. `air_temperature:*`, `air_temperature:*:mean:PT1M` or 'air_temperature:*:*:PT10M' etc. This also lets the user know if they accidentally write for example `air_temperature*:2.0:mean:PT1M`

api/utilities.py
```diff
@@ -71,9 +71,13 @@ async def verify_parameter_names(parameter_names: list) -> None:
...
 
     for i in parameter_names:
-        if i not in await get_current_parameter_names():
+        # HACK: This checking breaks the wildcard support (e.g. air_temperature:*:*:*). Skip if contains wildcards
+        if "*" in i:
+            continue
+        if i not in current_parameter_names:
```
fjugipe, line 80:
```suggestion
        parameter_name = i.split(':')[0]
        if parameter_name not in standard_names:
```

api/utilities.py
```diff
@@ -71,9 +71,13 @@ async def verify_parameter_names(parameter_names: list) -> None:
     Raises error with unknown names if any are found.
     """
     unknown_parameter_names = []
+    current_parameter_names = await get_current_parameter_names()
```
fjugipe, line 74:
```suggestion
    current_parameter_names = await get_current_parameter_names()
    standard_names = set(map(lambda x: x.split(":")[0], list(current_parameter_names)))
```

api/utilities.py
lukas-phaf: Personally, I wouldn't have a problem with just dropping the whole `verify_parameter_names` function.
Even if we keep it, I am okay with wildcards overriding it. If you know use wildcards, you should know what you are doing.

Haven't looked in detail, but with your suggestion, you only check the standard_name, correct? So `air_temperature:foo:bar:foo` now gives "data not found error".

I think something like `*:2.0:*:*` might be valid use case as well.

fjugipe: Yes, `air_temperature:foo:bar:foo` will return a "data not found error" with the proposed changes. But is that really a case, should we return a better error with that?

for `*:2.0:*:*` I think we could modify the check to: `if parameter_name not in standard_names and parameter_name != "*":` Which will also then allow `parameter-name=*`, but `parameter-name=foobar*` will return `Unknown parameter-name ['foobar*']` instead of "data not found error".

I made the proposes because I thought that the incorrect parameter-name is the most common use case. This check will also work if we do some modifications and allow "wildcard searches" with only using e.g. `parameter-name=air_temperature, air_pressure` etc. without the user having to type the '*' behind it.

Dropping the function altogether is a valid option 😎

fjugipe: Just noticed that the proposed changes won't allow wildcarding the parameter-name itself e.g. `air_*:*:mean:PT1M`. Which I think is a case, so I guess going with the `If you use wildcards, you should know what you are doing.` and keep the simplest solution at least for now 👍

fjugipe (approved):

api/utilities.py
lukas-phaf: I agree that we should review this (and will make an issue for that), but I will merge what we have now.

PR #129, opened 2024-06-04
Set up max connections for SQL connection pool.

lukas-phaf:
Double check this doesn't break anything by looking if the load test still works correctly.

Are we properly "closing" all connections, meaning giving them back to the pool, after usage?
Because it surprises me we are actually using all 100 allowed connections with the current setup, are we leaking them?

### DIFF
--- a/datastore/datastore/go.mod
+++ b/datastore/datastore/go.mod
@@ -2,18 +2,18 @@ module datastore
 
 go 1.20
 
-require google.golang.org/grpc v1.56.1
+require google.golang.org/grpc v1.64.0
 
 require (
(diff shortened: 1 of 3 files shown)


lukas-phaf:
Double check this doesn't break anything by looking if the load test still works correctly.

Are we properly "closing" all connections, meaning giving them back to the pool, after usage?
Because it surprises me we are actually using all 100 allowed connections with the current setup, are we leaking them?

jo-asplin-met-no:
Looks good.

jo-asplin-met-no (approved):
Looks good.

PR #135, opened 2024-06-17
Disable uniqe name constraint, and pick one

Teddy-1000:
Disable the unuqie name constriant and just pick a name. Name are not as important as station ID.

### DIFF
--- a/api/routers/edr.py
+++ b/api/routers/edr.py
@@ -134,13 +134,16 @@ async def get_locations(
                 },
             )
         if len(platform_names[station_id]) > 1:
-            raise HTTPException(
-                status_code=500,
-                detail={
-                    "platform_name": f"Station with id `{station_id} "
-                    f"has multiple names: {platform_names[station_id]}"
-                },
-            )
+            # Just pick the first name in the set, names can be changed, but stationID is static
+            platform_names[station_id] = platform_names[station_id].pop()
+
+            # raise HTTPException(
+            #    status_code=500,
+            #    detail={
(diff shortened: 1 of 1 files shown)


api/routers/edr.py
```diff
@@ -134,13 +134,16 @@ async def get_locations(
...
+            # Just pick the first name in the set, names can be changed, but stationID is static
+            platform_names[station_id] = platform_names[station_id].pop()
+
+            # raise HTTPException(
+            #    status_code=500,
+            #    detail={
+            #        "platform_name": f"Station with id `{station_id} "
+            #        f"has multiple names: {platform_names[station_id]}"
+            #    },
+            # )
```
lukas-phaf, line 146: This code feels fishy. Replace a list by a single element, and `pop` take a random element from the set, so not reproducible.

Why not just comment the whole `if`? Line 153 will take the first one anyway.
But probably best to change that into the following, so we always get a consistent answer (making a set into a list is also random):
`sorted(list(platform_names[station_id]))[0]`

Can you also add a TODO?

Teddy-1000: Yes, we can make it more reproducable. What would you say should be in the in the TODO?

lukas-phaf: Sorry, I mean add the sorted to line 153, and comment out the whole if statement.

lukas-phaf: On the TODO, something like: Do we want to check for duplicate names? Can we communicate this to the user without throwing an error?

PR #136, opened 2024-06-18
Issue 130 add unit parsing

Teddy-1000:
Add unit parsing and mapping file.


ingest/api/std_name_units.json
```diff
@@ -0,0 +1,424 @@
...
+            "Pa": {
+                "mul": 0.01
+            }
+        }
+    },
+    "air_temperature": {
+        "unit": "degC",
+        "conversion": {
+            "K": {
+                "add": 272.14
```
lukas-phaf, line 406: Pretty sure that:
Celcius = K - 273.15

So it should be `273.15` or `-273.15`, depending on which direction the code works.. Anyway, 0 C should come out as 273.15 K.

This is used in many places in this file.

Teddy-1000: It should indeed be `-273.15`, or it would be rather hot from the places reporting kelvins. Should be fixed now.

lukas-phaf: You still have `-273.14` in the code, should be `-273.15`.

istvans-mn: One digit can be enough. See [Table 1](https://library.wmo.int/viewer/68695/download?file=8_I-2023_en.pdf&type=pdf&navigator=1) , page 103.

lukas-phaf: The table is about measurement accuracy. I would say the conversion has to be as accurate as possible (-273.15 is exact). Because if the original measurement has 2 decimal places, the conversion would otherwise be wrong.

lukas-phaf: I don't think we currently control output number accuracy in the API. We give back whatever the original number that was put into the datastore by the ingestor.

istvans-mn: This is an [example](https://api.esoh.met.no/collections/observations/locations/0-578-0-46800?parameter-name=air_temperature%3A%2A%3A%2A%3A%2A&datetime=2024-06-21T03%3A00%3A00Z&f=CoverageJSON) :

`  "ranges": {
    "air_temperature:0.0:point:PT0S": {
      "type": "NdArray",
      "dataType": "float",
      "axisNames": [
        "t",
        "y",
        "x"
      ],
      "shape": [
        1,
        1,
        1
      ],
      "values": [
        9.19999981
      ]
    }
  }
`
Round to two digits?

ingest/api/std_name_units.json
```diff
@@ -0,0 +1,481 @@
...
+        "symbol": "degrees",
+        "alias": [
+            "deg",
+            "degree"
+        ]
+    },
+    "low_type_cloud_area_fraction": {
+        "unit": "octas",
+        "symbol": "octas"
+    },
```
fjugipe, line 59:
```suggestion
    "low_type_cloud_area_fraction": {
        "unit": "oktas",
        "symbol": "oktas",
        "alias": [
            "octas"
        ]
    },
```

https://en.wikipedia.org/wiki/Okta

ingest/api/std_name_units.json
```diff
@@ -0,0 +1,481 @@
...
+        "conversion": {
+            "K": {
+                "add": -273.15
+            }
+        }
+    },
+    "cloud_area_fraction": {
+        "unit": "octas",
+        "alias": "octa"
+    },
```
fjugipe, line 469:
```suggestion
    "cloud_area_fraction": {
        "unit": "oktas",
        "symbol": "oktas",
        "alias": [
            "okta",
            "octa",
            "octas"
        ]
    },
```

https://en.wikipedia.org/wiki/Okta

ingest/api/std_name_units.json
```diff
@@ -0,0 +1,481 @@
...
+    "duration_of_sunshine": {
+        "unit": "s",
+        "symbol": "s",
+        "conversion": {
+            "minutes": {
+                "mul": 60
+            }
+        }
+    },
```
fjugipe, line 26:
```suggestion
    "duration_of_sunshine": {
        "unit": "s",
        "symbol": "s",
        "conversion": {
            "minutes": {
                "mul": 60
            },
            "h": {
                "mul": 3600
            }
        }
    },
```

ingest/api/std_name_units.json
Teddy-1000: I have updated the code to keep as many decimals as the original measurement.

shamlymajeed (approved):

PR #142, opened 2024-06-27
(WIP) Improve EDR query functionality to make it easier to get "all temperatures for Europe"

lukas-phaf:
Issue #140

Issue #140, opened 2024-06-26
[API] Improve EDR query functionality to make it easier to get "all temperatures for Europe"


lukas-phaf:
The EDR parameter names currently consist of four parts, an example is `air_temperature:1.5:mean:PT1M`.
So the parts are: `standard_name:level:function:period`.

Even though wildcards are allowed when querying, e.g. `air_temperature:*:*:*`, both the query side and response side of things could be improved. 

My suggestion is to keep the parameter names (for now), but make them less important for querying and understanding the data. It should be possible to use the data without parsing/understanding the parameter names.

Make the following changes to the query side:
- Introduce query parameter `z`, which can be used to query for level (including range `z_1/z_2`)
- Introduce custom dimensions `standard_name` (enum), `function` (enum) and `period` (enum for now, but can be range later)
- Do the `z` filtering in the API for now, can later be moved to the datastore

An example query to get all air temperatures for Finland at level between 1.25 and 2 meters in a single query:
`https://api.esoh.met.no/collections/observations/area?coords=POLYGON%20((28.59193%2069.064777,%2028.445944%2068.364613,%2029.977426%2067.698297,%2029.054589%2066.944286,%2030.21765%2065.80598,%2029.54443%2064.948672,%2030.444685%2064.204453,%2030.035872%2063.552814,%2031.516092%2062.867687,%2031.139991%2062.357693,%2030.211107%2061.780028,%2028.069998%2060.503517,%2026.255173%2060.423961,%2024.496624%2060.057316,%2022.869695%2059.846373,%2022.290764%2060.391921,%2021.322244%2060.72017,%2021.544866%2061.705329,%2021.059211%2062.607393,%2021.536029%2063.189735,%2022.442744%2063.81781,%2024.730512%2064.902344,%2025.398068%2065.111427,%2025.294043%2065.534346,%2023.903379%2066.006927,%2023.56588%2066.396051,%2023.539473%2067.936009,%2021.978535%2068.616846,%2020.645593%2069.106247,%2021.244936%2069.370443,%2022.356238%2068.841741,%2023.66205%2068.891247,%2024.735679%2068.649557,%2025.689213%2069.092114,%2026.179622%2069.825299,%2027.732292%2070.164193,%2029.015573%2069.766491,%2028.59193%2069.064777))&standard_name=air_temperature&z=1.25/2`

Response side:
[...]

PR #146, opened 2024-07-18
(WIP) Improve EDR query functionality to make it easier to get "all temperatures for Europe", try 2

lukas-phaf:
Issue #140.

Compared to PR #142, in this PR we don't implement `level` as a `z` paremeter, instead it is just another 'custom dimension'.

TODO:
- [x] Fix unit tests
- [x] Add unit/integration tests for custom dimensions `standard_names`, `levels`, `functions` and `periods`.
- [x] Fix some of the TODO's in the code.
- [ ] Deploy before merging so Mikko can confirm that this actually improved the API functionality?


api/formatters/covjson.py
```diff
@@ -28,7 +29,15 @@
...
 def make_parameter(ts_mdata):
+    custom_fields = {
+        "rodeo:standard_name": ts_mdata.standard_name,
+        "rodeo:level": float(ts_mdata.level) if is_float(ts_mdata.level) else 0.0,
+        "rodeo:function": ts_mdata.function,
+        "rodeo:period": ts_mdata.period,
+    }
+
     return Parameter(
+        # TODO: Change description, as we know have the explicit fields?
```
fjugipe, line 40: In my opinion for example: `air_pressure at 1.5m PT1M mean` is a good description of the parameter and could be kept as such.

fjugipe:
Some TODO's have been left open and most features require refactoring after levels and periods are stored as integers in the datastore as stated in #153 and #155.

After this merge it is possible to filter observations with several custom dimensions with all `/collections` data queries such as:
1. **standard_names**
   -  List of `standard_names` such as: `/collections/observations/locations?standard_names=air_temperature, wind_speed` to get all available locations with all different `air_temperature` and `wind_speed` observations
2. **levels**
   - List of `levels` such as: `/collections/observations/position?coords=POINT(5.179705 52.0988218)&levels=1.0, 2.0` to get all available observations with levels of 1.0m and 2.0m
   - Range of `levels` such as: `/collections/observations/position?coords=POINT(5.179705 52.0988218)&levels=1.0/10.0` to get all available observations between the levels of 1.0m and 10.0m
   - Interval of `levels` such as: `/collections/observations/position?coords=POINT(5.179705 52.0988218)&levels=R5/0.5/0.5` to get all available observations from 5 steps of levels, starting from 0.5m with increases of 0.5m as in 0.5m, 1.0m, 1.5m, 2.0m and 2.5m
3. **periods**:
   - List of `periods` such as `/collections/observations/area?coords=POLYGON((5.0 52.0, 6.0 52.0,6.0 52.1,5.0 52.1, 5.0 52.0))&periods=PT0S, PT1M, PT10M`
   - Range of `periods` such as `/collections/observations/area?coords=POLYGON((5.0 52.0, 6.0 52.0,6.0 52.1,5.0 52.1, 5.0 52.0))&periods=PT0S/PT10M`
4. **functions**:
   -  List of functions `/collections/observations/locations/{location_id}?functions=mean,maximum`
5. **combination**
   -  Combination of different custom filters above such as `/collections/observations/locations/{location_id}?functions=mean,maximum&periods=PT0S/PT01H&standard_names=air_temperature`
   
Ranges on `periods` and `levels` allow wildcarding the start and/or end of the ranges with `..` such as `ranges=../10.0` or `periods=PT0S/..`

lukas-phaf:
The openapi examples setup is really nice!

api/utilities.py
```diff
@@ -137,3 +147,159 @@ async def add_parameter_name_and_datetime(request, parameter_name: str | None, d
...
+async def get_unique_values_for_metadata(field: str) -> list[str]:
+    request = dstore.GetTSAGRequest(attrs=[field])
+    response = await get_ts_ag_request(request)
+    return [getattr(i.combo, field) for i in response.groups]
+
+
+def filter_observations_for_z(observations, z):
+    if z:
+        z_values = get_z_levels_or_range(z)
+        if len(z.split("/")) == 2:
```
lukas-phaf, line 230: Maybe better to make `get_z_levels_or_range()` return a tuple if it is a range, and a list if list of valid values, and then `if` on the type? It feels a bit strange to do the `split` here again to make the decision?

api/utilities.py
```diff
@@ -137,3 +147,159 @@ async def add_parameter_name_and_datetime(request, parameter_name: str | None, d
...
+    # Make sure the pattern is Rn/n/n
+    # Allow optional decimals in interval and starting value
+    # TODO: Can the starting value be negative? Can the interval be negative to go in reverse?
+    pattern = re.compile(r"^R\d+/\d+(\.\d+)?/\d+(\.\d+)?$")
+    if not pattern.match("/".join(interval)):
+        raise HTTPException(status_code=400, detail=f"Invalid levels repeating-interval: {'/'.join(interval)}")
+
+    amount_of_intervals = int(interval[0][1:])
+    interval_value = float(interval[1])
+    starting_value = float(interval[2])
```
lukas-phaf, line 205: Not consistent with line 177. Not sure which is correct.

api/utilities.py
```diff
@@ -137,3 +147,159 @@ async def add_parameter_name_and_datetime(request, parameter_name: str | None, d
...
+
+
+def numeric_sort_key(value: str) -> float:
+    """
+    Converts a string to a float for comparison, returns infinity if the string is not convertible.
+    """
+    try:
+        return float(value)
+    except ValueError:
+        return float("inf")
```
lukas-phaf, line 251: Would NaN be better, or do you specifically need something big?

fjugipe: I was under the impression that using NaN in sorting could have some unexpected behavior, so I decided to use positive `inf` to push non float values to the end of the list. The ingestor *should* take care that there are no `levels` in the system which are not castable to `float`, so this was just a *"better safe than sorry"* solution.

api/routers/edr.py
```diff
@@ -54,21 +56,58 @@
...
+        str | None,
+        Query(
+            description="Comma seperated list of parameter aggregation methods",
+            openapi_examples=custom_dimension_examples.methods,
+        ),
+    ] = None,
+    periods: Annotated[
+        str | None,
+        Query(
+            description="Comma seperated list of parameter aggregation periods",
```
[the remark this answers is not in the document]
fjugipe: Improved all of the former in https://github.com/EURODEO/e-soh/pull/146/commits/4a6de7488e3c35b4d734b05ede60da828d9d7c83, used same logic as with the openapi examples so they are easier to maintain.

api/formatters/covjson.py
fjugipe: Fixed in: https://github.com/EURODEO/e-soh/pull/146/commits/19550f0927d6e09cca51a838c6375284245390c9

api/utilities.py
```diff
@@ -137,3 +147,159 @@ async def add_parameter_name_and_datetime(request, parameter_name: str | None, d
...
+        else:
+            end_index = len(periods) - 1
+
+        if start_index > end_index:
+            raise HTTPException(status_code=400, detail=f"Invalid ISO 8601 range: {start} > {end}")
+
+    except ValueError as err:
+        raise HTTPException(status_code=400, detail=f"Invalid ISO 8601 range: {err} of possible period values")
+
+    return periods[start_index : end_index + 1]  # noqa: E203
```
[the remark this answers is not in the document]
fjugipe: Better range support in: https://github.com/EURODEO/e-soh/pull/146/commits/f1196f886182413fc635e5b621296691c5820e29

Decided to make a separate function for transforming durations to seconds, this function is in line with what @Teddy-1000 did in the upcoming ingest change. The sorting functions purpose is to sort, so I kept it as such. It will be obsolete anyways after storing periods as seconds in the database.

api/utilities.py
fjugipe: fixed in: https://github.com/EURODEO/e-soh/pull/146/commits/f52ece18e197057a552aba7f4908f3f500c3420b

api/utilities.py
fjugipe: Somewhere in the making they went into wrong order, but fixed in: https://github.com/EURODEO/e-soh/pull/146/commits/a9fcc8d3d1f51852b059081298641c22e9d69569

api/utilities.py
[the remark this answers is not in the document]
lukas-phaf: I am confused... the `iso_8601_duration_to_seconds_sort_key` function does exactly the same as `iso_8601_duration_to_seconds`, so why keep it?

This looks pretty clear to me:
`sorted(await get_unique_values_for_metadata("period"), key=iso_8601_duration_to_seconds)`

If you really want to keep the `iso_8601_duration_to_seconds_sort_key`, just call the other function from there.
Besides the maintenance, having two functions which try to do the same thing in a different way might give some really confusing results...

Or am I missing something?

api/routers/edr.py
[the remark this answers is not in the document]
lukas-phaf: Nice.

api/utilities.py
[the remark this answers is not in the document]
fjugipe: Almost exactly the same. Other one is used to `sort` durations inside the system and it will not hinder the system unusable and raise an error if the system for some reason contains data which cannot be sorted, it handles it "gracefully" and just pushes them at the end of the list. We can get rid of it after everything is stored as integers, since they are always sortable.

The other one is used to transform durations to seconds, but also `validate` user queries and will raise an error if it finds a duration which cannot be transformed to seconds. 

Using the former for example with sorting the periods for `/collections` with unsortable data would result it raising an error and I don't think that is how custom sorting algorithm should work.

api/utilities.py
```diff
@@ -137,3 +149,179 @@ async def add_parameter_name_and_datetime(request, parameter_name: str | None, d
...
+        # If the duration string is not in the correct format, return the maximum integer value
+        return sys.maxsize
+
+    days = int(match.group(1) or 0)
+    hours = int(match.group(2) or 0)
+    minutes = int(match.group(3) or 0)
+    seconds = int(match.group(4) or 0)
+
+    total_seconds = days * 86400 + hours * 3600 + minutes * 60 + seconds
+    return total_seconds
```
lukas-phaf, line 276:
```suggestion
def iso_8601_duration_to_seconds_sort_key(duration: str) -> int:
    try:
        seconds = iso_8601_duration_to_seconds(duration)
    except ValueError:
        seconds = sys.maxsize
    return seconds
```

api/utilities.py
```diff
@@ -137,3 +149,164 @@ async def add_parameter_name_and_datetime(request, parameter_name: str | None, d
...
+        seconds_in_seconds = duration.tdelta.seconds
+        total_seconds = years_in_seconds + months_in_seconds + days_in_seconds + seconds_in_seconds
+    else:
+        # It's a simple timedelta, so just get the total seconds
+        total_seconds = duration.total_seconds()
+
+    return int(total_seconds)
+
+
+async def get_iso_8601_range(start: str, end: str) -> list[str] | None:
```
lukas-phaf, line 284: Found a bug in this. If the requested range is empty ('PT7H/PT7H'), no filtering on period is done.
The issue is that an empty list is then passed to the Datastore filtering, which means no filtering on that parameter.

Maybe best to throw and error directly from this function (like for start > end)?
And add a test for it, to avoid regressions?

fjugipe: Fixed in: https://github.com/EURODEO/e-soh/pull/146/commits/8ef434ce941fe37c565e4872ff6da5028d24d912

api/routers/edr.py
```diff
@@ -25,7 +28,7 @@
 from shapely import geometry
 from shapely import wkt
 from shapely.errors import GEOSException
-from utilities import add_parameter_name_and_datetime
+from utilities import add_request_parameters, filter_observations_for_z
```
Teddy-1000, line 31:
```suggestion
from utilities import add_request_parameters
from utilities import filter_observations_for_z
```

Teddy-1000 (approved):

PR #170, opened 2024-09-16
Add metrics endpoint to edr api

Teddy-1000:


api/Dockerfile
```diff
@@ -28,6 +28,9 @@ RUN python -m grpc_tools.protoc  \
```
Jeffrey-Vervoort-KNMI, line 28: I randomly checked some of the metrics. The numbers look good. I think it contains a lot of important metrics such as: request-size, -sum, -total, -duration and the mean can be calculated. And all of those are available per endpoint. There are two metrics that I think we could consider to add. What do you think about a metric for:

- Uptime Gauge
- HTTP status code counter (For an overview of errors, timeouts and successful requests)

api/requirements.txt
```diff
@@ -1,8 +1,8 @@
 #
-# This file is autogenerated by pip-compile with Python 3.11
+# This file is autogenerated by pip-compile with Python 3.10
```
Jeffrey-Vervoort-KNMI, line 2: Can you compile it with 3.11 for consistency with the docker image?

api/Dockerfile
Teddy-1000: Uptime need to be handled by the host machine/VM.
HTTP status codes are already provided, but do not appear until a request with the status code is given. Also 2xx, 3xx and 5xx are not split but default.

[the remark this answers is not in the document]
Teddy-1000: The metrics will be scrapable from /metrics endpoint.

Jeffrey-Vervoort-KNMI (approved):

PR #181, opened 2024-09-20
Issue 178 levels filtering with multiple ranges

fjugipe:
- Allow using multiple ranges and multiple repeating-intervals for filtering data with `levels` custom dimension
- Refactor `periods` custom dimension range filtering to use datastores open range filtering for integers


fjugipe:
Check #214 instead

PR #182, opened 2024-09-20
fix level and period for mqtt

shamlymajeed:


ingest/api/model.py
```diff
@@ -343,7 +344,7 @@ def transform_period_to_seconds(self):
         else:
             raise ValueError("Duration not convertable to seconds.")
 
-        self.period = int(total_seconds)
+        self.period_convertable = int(total_seconds)
```
Teddy-1000, line 347: Could we maybe call this period_integer? So we can se what is it?

Teddy-1000 (approved):

PR #184, opened 2024-09-20
Removed unreferenced time series and geo points

jo-asplin-met-no:


jo-asplin-met-no:
> Can we decouple this from the `merge-backend-duration-and-level-changes` branch, and merge it directly to main?

Fine by me, since i guess this is unrelated to that merge-backend... task anyway. But what's the best way to do that?

jo-asplin-met-no:
Maybe the easiest is to create an altogether new branch (based off main) + PR ?

jo-asplin-met-no:
Another solution is maybe to just merge into merge-backend-... as planned, but in addition create a new branch/PR for getting the relevant commits into main? As you can tell, I don't feel proficient enough in git to see what would be the best solution.

lukas-phaf:
I am not a git expert either. I tried making a merge request for your branch against master, but that gives a huge amount of changes, I think because the `merge-backend-duration-and-level-changes` are your branch are behind main. Rebasing on main of your branch might work, but it is probably a nightmare.

Personally, I would create patch from this PR, and apply it to master to create a new PR (so what you suggest above).

@Jeffrey-Vervoort-KNMI, do you have a suggestion how to handle this? This is currently a branch of a branch, but the code is unrelated to that branch, and it would be better if it goes straight into main.

jo-asplin-met-no:
Hm, I think the best is simply to start from scratch, i.e. create a new branch off main, add a single commit with the net change, and then a PR for that. Stay tuned ...

jo-asplin-met-no:
Done! Forget this PR and use [this one](https://github.com/EURODEO/e-soh/pull/192) instead!

PR #194, opened 2024-10-01
Build images when creating tags

Teddy-1000:


.github/workflows/registry_upload.yml
```diff
@@ -68,6 +71,8 @@ jobs:
           tags: |
             type=raw,value=latest,enable=${{ github.ref == format('refs/heads/{0}', 'main') }}
             type=ref,event=tag,enable={{is_default_branch}}
+            type=semver,pattern={{version}}
```
lukas-phaf, line 74: Looking at the documentation, this is duplicate. Or actually, it wil tag with both `v0.2.0` and `0.2.0`.

lukas-phaf: Looking at [this](https://github.com/docker/metadata-action?tab=readme-ov-file#latest-tag), we just need the semvar line, together with `flavor latest=auto`?

.github/workflows/registry_upload.yml
```diff
@@ -7,6 +7,9 @@ on:
     branches: [main]
     types:
       - completed
```
lukas-phaf, line 9: Should we drop the `workflow_run` bit? I think you would expect `latest` to point to the latest `release`, not just the latest commit?

.github/workflows/registry_upload.yml
lukas-phaf: Although that means if you tag a test branch, it will also be tagged latest, which is not good. So maybe this?

```
          tags: |
            type=raw,value=latest,enable={{is_default_branch}}
            type=semver,pattern={{version}}
```

lukas-phaf: Hmm, `semver` would not trigger on a tag `test-123`, so that is not possible anyway. But it would build, just not get tagged? This is pretty complicated.

Teddy-1000: I will change to only run on tags, from main it will only tag if the tag is a semver, if you a a commit on any other branch it will build images with that exact tag.

.github/workflows/registry_upload.yml
Teddy-1000: Yes, we can drop the workflow_run.

lukas-phaf (approved):

PR #198, opened 2024-10-02
Create container for migrate

### LINKED ISSUE #187: Build and publish container for `migrate`
We need to build our own migrate container with the following changes:
- bake the migration scripts into it (they are currently mounted as a volume)
- Allow control of db host, database name, user and password through environment variable

Teddy-1000:
Created a dockerfile for the migrate scripts instead of mounting them.
Closes #187


datastore/migrate/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM migrate/migrate:4
+
+COPY data/migrations /migrations
+
+COPY ./run-migrate.sh /
+
+RUN chmod +x /run-migrate.sh
```
lukas-phaf, line 7: This line is not needed, as `run-migrate.sh` is marked `+x` in the repo. Looks like the `COPY` keeps that.

Teddy-1000: Ah, yes. Will fix it, originally it was not, but the pre-commit complained when a file with a shebang was not executable.

lukas-phaf (approved):

PR #199, opened 2024-10-02
Restore lost code

Teddy-1000:


protobuf/datastore.proto
```diff
@@ -305,15 +305,9 @@ message GetTSAGResponse {
...
-
-  // general filter
-  // - map keys must correspond exactly with string field names in TSMetadata or ObsMetadata
-  // - if map key F is specified (where F is for example 'platform'), only observations that match
-  //   at least one of these values for F may be returned
-  map<string, Strings> filter = 1;
+  // No parameters
 
-  // --- END filter for reflectable metadata of type int64 or string -------------------------
+  // --- END string metadata (handleable with reflection) -------------------------
```
lukas-phaf, line 310: This line should be deleted, only now see why it was there.

lukas-phaf (approved):

PR #206, opened 2024-10-10
Update landing page info

Teddy-1000:
PR for updating landing page info.

### DIFF
--- a/api/openapi/custom_dimension_examples.py
+++ b/api/openapi/custom_dimension_examples.py
@@ -36,6 +36,10 @@ levels = {
         "summary": "Open start range",
         "value": "../10.0",
     },
+    "Combination": {
+        "summary": "Combination",
+        "value": "1.0, 1.5/1.8, R5/2.0/2.0",
+    },
 }
 
 periods = {
@@ -53,6 +57,10 @@ periods = {
     "Wildcard": {
         "summary": "Open end range",
         "value": "PT0S/..",
+    },
+    "Combination": {
+        "summary": "Combination",
+        "value": "PT0S, PT1M/PT10M",
     },
 }
 
--- a/api/test/test_api_utilities.py
+++ b/api/test/test_api_utilities.py
@@ -1,0 +1,19 @@
+import pytest
+
+from utilities import get_levels_values
+
+levels_in_out = [
+    ("1", ["100"]),
+    ("1,2", ["100", "200"]),
+    ("1,2, 3", ["100", "200", "300"]),
+    ("1/3", ["100/300"]),
+    ("../3", ["../300"]),
+    ("1/..", ["100/.."]),
+    ("R3/1.2/0.3", ["120", "150", "180"]),
+    ("1, 3/5, R3/5/0.1,11", ["100", "300/500", "500", "510", "520", "1100"]),
+]
+
+
+@pytest.mark.parametrize("levels_in, levels_out", levels_in_out)
+def test_get_levels_values(levels_in, levels_out):
+    assert get_levels_values(levels_in) == levels_out
--- a/api/test/test_edr.py
+++ b/api/test/test_edr.py
@@ -410,3 +410,33 @@ def test_get_data_with_lowercase_period_range_without_existing_data():
 
         assert response.status_code == 404
         assert response.json() == {"detail": "Requested data not found."}
+
+
+def test_get_data_with_combination_levels_filtering():
+    with patch("routers.edr.get_obs_request") as mock_get_obs_request:
+
+        # Load with random test data for making a mock_obs_request
+        test_data = load_json("test/test_data/test_coverages_proto.json")
+        mock_get_obs_request.return_value = create_mock_obs_response(test_data)
+
+        response = client.get("/collections/observations/locations/0-20000-0-06260?levels=R6/0.0/0.1, 1.5/1.8, 10")
+
+        m_args = mock_get_obs_request.call_args[0][0]
+
+        assert response.status_code == 200
+        assert m_args.filter["level"].values == ["0", "10", "20", "30", "40", "50", "150/180", "1000"]
+
+
+def test_get_data_with_combination_periods_filtering():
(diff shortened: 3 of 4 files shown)


api/openapi/openapi_metadata.py
```diff
@@ -0,0 +1,31 @@
+openapi_metadata = {
+    "title": "Observations Europe EUMETNET",
+    "description": (
+        "EDR data service for European environmental observations from EUMETNET," " co-funded by the European Union"
```
lukas-phaf, line 4:
```suggestion
        "EDR data service for European environmental observations from EUMETNET," " co-funded by the European Union."
```

lukas-phaf, line 4: Isn't "environmental observations" to broad?

api/openapi/openapi_metadata.py
```diff
@@ -0,0 +1,31 @@
...
+    ),
+    "contacts": [
```
lukas-phaf, line 6: According to [this](https://fastapi.tiangolo.com/reference/fastapi/#fastapi.FastAPI), there is no `contacts` parameter, only `contact`.

It also doesn't show up anywhere in the output, as far as I can see.

api/openapi/openapi_metadata.py
```diff
@@ -0,0 +1,31 @@
...
+            "links": [
+                {
+                    "rel": "about",
+                    "type": "text/html",
+                    "href": "https://www.eumetnet.eu/about-us/",
+                }
+            ],
+            "roles": ["host"],
+        },
+    ],
```
lukas-phaf, line 26: What use is a postal address to the user of this API?
Also, in the landingpage we put a different address (rodeoproject@fmi.fi), which is probably much more useful to the users...

api/metadata_endpoints.py
```diff
@@ -26,7 +26,11 @@
 from grpc_getter import get_ts_ag_request
 
 import datastore_pb2 as dstore
-from utilities import get_unique_values_for_metadata, seconds_to_iso_8601_duration, convert_cm_to_m
+from utilities import (
+    get_unique_values_for_metadata,
+    seconds_to_iso_8601_duration,
+    convert_cm_to_m,
+)
```
lukas-phaf, line 33: I would prefer to avoid this syntax. I thought we had linting that requires each to be on a separate line?

api/metadata_endpoints.py
```diff
@@ -56,14 +60,34 @@ def datetime_to_iso_string(value: datetime) -> str:
 
 def get_landing_page(request):
     return LandingPageModel(
-        title="E-SOH EDR API",
-        description="The E-SOH EDR API",
-        keywords=["weather", "temperature", "wind", "humidity", "pressure", "clouds", "radiation"],
+        title="Observations Europe EUMETNET",
+        description=(
+            "EDR data service for European environmental observations from EUMETNET," " co-funded by the European Union"
```
lukas-phaf, line 65: This is now duplicated.

lukas-phaf, line 65: I wasn't part of this discussion, but I think it should be in the title that it is an EDR API.

Also, why is it now called an "EDR data service", and not an EDR API?
I would even refer to OGC EDR API here.

api/metadata_endpoints.py
```diff
@@ -120,7 +148,7 @@ async def get_collection_metadata(base_url: str, is_self) -> Collection:
             description=f"{ts.standard_name} at {level}m, aggregated over {period} with method '{ts.function}'",
             observedProperty=ObservedProperty(
                 id=f"https://vocab.nerc.ac.uk/standard_name/{ts.standard_name}",
-                label=ts.parameter_name,
+                label=ts.standard_name,
```
lukas-phaf, line 151: Can you change all the `label`'s in the `ObservedProperty` to standard name? There is at least one in the CoverageJSON output.

datastore/integration-test/response/metadata_collections_all_collections.json
```diff
@@ -189,73 +189,73 @@
                     "rodeo:standard_name": "air_pressure_at_sea_level",
                     "rodeo:level": 1.0
                 },
-                "air_temperature:0.1:minimum:PT10M": {
+                "air_temperature:0.0:minimum:PT12H": {
```
lukas-phaf, line 192: Any idea why the order changed?

api/metadata_endpoints.py
Teddy-1000: I also thought we had that. Strange, will fix it.

datastore/integration-test/response/metadata_collections_all_collections.json
Teddy-1000: Best guess is that it is now sorted and was not in the past. I don't think the diff cares about the order of keys.

api/openapi/openapi_metadata.py
Teddy-1000: Good question. It is the only type of observations we have. I don't really have any better suggestions either.

lukas-phaf: `... European meteorological observations ...`

api/metadata_endpoints.py
```diff
@@ -56,14 +60,32 @@ def datetime_to_iso_string(value: datetime) -> str:
...
+            "weather",
+            "temperature",
+            "wind",
+            "humidity",
+            "pressure",
+            "clouds",
+            "radiation",
+        ],
         provider=Provider(name="RODEO", url="https://rodeo-project.eu/"),
         contact=Contact(email="rodeoproject@fmi.fi"),
```
lukas-phaf, line 75: Do we want this contact to be different from the one at API level?

api/openapi/openapi_metadata.py
```diff
@@ -0,0 +1,15 @@
+openapi_metadata = {
+    "title": "Observations Europe EUMETNET",
```
lukas-phaf, line 2: I still don't like this title. It should say "EDR" and "API".

PR #209, opened 2024-10-18
ObservedProperty label & test data loader through ingestion

### LINKED ISSUE #190: [TESTING] Re-write the knmi-client-station.py to use the ingest API.

### LINKED ISSUE #204: [API] Set ObservedPropery label to standard_name
I think the label in ObservedProperty should be change to the standard name.

fjugipe:
- Change `ObservedProperty` label to `standard_name` without underscores
- Change `Parameter` description to have `standard_name` without underscores
- Update test data to be on par with current ingest
- Add a data loader to load test data through ingest
- Add Dockerized unit tests for ingest
- Update `ci.yml` to load test data through ingest and 
- Comment test coverage for ingest unit tests
- Add new `just` commands

closes #204, closes #190


lukas-phaf:
It is not clear to me why `test-ingest` is a separate job. I think it leads to a confusing mix of jobs and steps in the action output.

Also, the matrix of python versions leads to additional confusion, especially as it has only one version. Can we take the blue pill and comment out the whole `strategy` bit for now so the matrix doesn't show up?

To be honest, I also don't see the advantage of a separate `publish-test-results` job. It was added by @Jeffrey-Vervoort-KNMI  as separation of concern. But it also adds a lot of noise (for example to retrieve the artifacts). I don't think the `contents` and `issues` permissions are [needed](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token), so I would be fine with all the main job having `pull-requests` permissions.

api/formatters/covjson.py
```diff
@@ -40,12 +41,11 @@ def make_parameter(ts_mdata):
 
     return Parameter(
         description={
-            "en": f"{ts_mdata.standard_name} at {level}m, "
-            f"aggregated over {period} with method '{ts_mdata.function}'",
+            "en": f"{label} at {level}m, " f"aggregated over {period} with method '{ts_mdata.function}'",
```
lukas-phaf, line 44: Make this one string.

datastore/data-loader/client_knmi_station_ingest.py
```diff
@@ -0,0 +1,254 @@
+#!/usr/bin/env python3
```
lukas-phaf, line 1: I guess comparing this with `client_knmi_station.py` making this work with one file would only make it more confusing.

But can we move some shared helper stuff to a shared file. Like everything above `netcdf_file_to_requests()` and also `generate_parameter_name()` (the only real difference seems to be the level, which can easily be translated after calling this function?

datastore/data-loader/Dockerfile
```diff
@@ -29,7 +30,16 @@ RUN python -m grpc_tools.protoc  \
...
+# Copy both loaders and conditionally use the wanted one
+COPY "./client_knmi_station_ingest.py" "${DOCKER_PATH}/client_knmi_station_ingest.py"
 COPY "./client_knmi_station.py" "${DOCKER_PATH}/client_knmi_station.py"
 
+RUN if [ "${THROUGH_INGEST}" = "true" ]; then \
+      mv "${DOCKER_PATH}/client_knmi_station_ingest.py" "${DOCKER_PATH}/client_knmi_station.py"; \
+    else \
+      rm "${DOCKER_PATH}/client_knmi_station_ingest.py"; \
+    fi
+
```
lukas-phaf, line 43: A bit nasty, but I don't see a better way.

datastore/data-loader/client_knmi_station_ingest.py
```diff
@@ -0,0 +1,254 @@
...
+    with Pool(cpu_count()) as pool:
+        results = pool.starmap(send_request_to_ingest, [(msg, url) for msg in observation_request_messages])
+
+        for status_code, response in results:
+            if status_code != 200:
+                print(status_code, response)
+                return
+
+
+def main(observation_request_messages: List, url):
```
lukas-phaf, line 234: Calling this `main` might be confusing. And I know, my bad for writing a bunch of code in:
`if __name__ == "__main__":`

Calling it `insert_data` would make sense, I think?

datastore/data-loader/client_knmi_station_ingest.py
```diff
@@ -0,0 +1,254 @@
...
+        response = requests.post(url, data=json.dumps(msg))
+        response.raise_for_status()
+        return response.status_code, response.json()
+    except requests.RequestException as e:
+        return response.status_code, e
+
+
+def insert_data(observation_request_messages: List, url):
+    with Pool(cpu_count()) as pool:
+        results = pool.starmap(send_request_to_ingest, [(msg, url) for msg in observation_request_messages])
```
lukas-phaf, line 226: Nice trick, didn't know about the `starchart` function.
I would have gone for `functools.partial` but this works as well. And it is good to know it exists. Although this is a slight abuse, I would say, as the `url` is fixed.

datastore/data-loader/client_knmi_station.py
```diff
@@ -29,6 +29,32 @@
...
+    standard_name_mapping = {
+        "cloud_cover": "cloud_area_fraction",
+        "total_downwelling_shortwave_flux_in_air": "surface_downwelling_shortwave_flux_in_air",
+        "precipitation_rate": "rainfall_rate",
+        "air_pressure_at_sea_level": "air_pressure_at_mean_sea_level",
+    }
+    return standard_name_mapping.get(standard_name, standard_name)
+
+
+# NOTE: Only units are converted currently, not values.
```
lukas-phaf, line 42: Definetly something to be aware of, as we know have "unlikely" data.

datastore/data-loader/requirements.txt
```diff
@@ -2,44 +2,54 @@
 # This file is autogenerated by pip-compile with Python 3.11
 # by the following command:
 #
-#    pip-compile --no-emit-index-url ./requirements.in
+#    pip-compile --no-emit-index-url
```
lukas-phaf, line 5: We keep getting pingpong with this. If you use `just pip-compile` you generate them all in a consistent way.
But maybe not something you want to deal with in this PR.

datastore/data-loader/client_knmi_station.py
```diff
@@ -29,6 +29,32 @@
...
+    unit_mapping = {
+        "degrees Celsius": "degC",
+        "ft": "m",
+        "min": "s",
+        "degree": "degrees",
+        "%": "percent",
+        "mm": "kg/m2",
+        "m s-1": "m/s",
+        "octa": "oktas",
+        "W m-2": "W/m2",
```
lukas-phaf, line 53: These are the one that the ingestor translates units to?

datastore/data-loader/client_knmi_station.py
```diff
@@ -58,42 +84,52 @@ def netcdf_file_to_requests(file_path: Path | str) -> Tuple[List, List]:
...
                     naming_authority="nl.knmi",
                     keywords=file["iso_dataset"].attrs["keyword"],
+                    summary=file["iso_dataset"].attrs["abstract"],
                     keywords_vocabulary=file.attrs["references"],
                     source=file.attrs["source"],
                     creator_name="KNMI",
                     creator_email=file["iso_dataset"].attrs["email_dataset"],
                     creator_url=file["iso_dataset"].attrs["url_metadata"],
-                    creator_type="Institution",
+                    creator_type="institution",
```
lukas-phaf, line 122: Is this actually an Enum in the ingestor?

fjugipe:
I see the point on separate jobs for tests when you trigger them for example `on push` to that `service`. Currently we don't have that, but should we? Running performance tests to the `api` on each push, when changes are made to `ingest` seems a bit unnecessary in my opinion and I personally would run them only `on pull_request`.

I agree with you on the changes, especially the `strategy` is unnecessary now that the `ingest` tests are run in a container. Overall I think that our `workflows` could have a slight rework. But in this PR I changed only minor things, since it is for another issue. 

I can continue refactoring the `workflows` here, but I think the work should be done in a separate branch, since this PR contains changes which affect how we present the data.

datastore/integration-test/test_knmi.py
```diff
@@ -46,7 +46,10 @@ def test_find_series_single_station_all_parameters(grpc_stub):
     request = dstore.GetObsRequest(filter=dict(platform=dstore.Strings(values=["0-20000-0-06260"])))
     response = grpc_stub.GetObservations(request)
 
-    assert len(response.observations) == 42  # Station 06260 doesn't have all parameters
+    # HACK: data-loader which straight to datastore doesn't drop the duplicates
+    assert (
+        len(set([obs.ts_mdata.parameter_name for obs in response.observations])) == 34
+    )  # Station 06260 doesn't have all parameters
```
lukas-phaf, line 52: Not clear to me why this is needed.

justfile
```diff
@@ -92,6 +95,8 @@ integration:
 unit:
     docker compose run --rm api-unit
 
+ingest_unit:
+    docker compose run --rm ingest-unit
```
lukas-phaf, line 99: So there were already ingest unit tests, they were just not part of `just`?

lukas-phaf (commented):
Great improvements!

Jeffrey-Vervoort-KNMI:
The main reason was minimal principle. The other parts of the jobs did not need write permission. Looking at the documentation again, I think that as Lukas says, the permissions for `issues` and `contents` can be removed.
________________________________

api/formatters/covjson.py
fjugipe: Fixed in: https://github.com/EURODEO/e-soh/pull/209/commits/167e77552c40480e7aaa3b9453c2e720c24a97a0

datastore/data-loader/client_knmi_station_ingest.py
fjugipe: Moved everything to `utilities.py` in https://github.com/EURODEO/e-soh/pull/209/commits/c5de77e91087ab4e7d9c9af0c438c406db00e4bc and doing some translations where needed in the `data-loaders`

datastore/data-loader/Dockerfile
fjugipe: Yeah, it is either this or two separate `Dockerfiles` to be maintained. It is ugly, but easier to main and this way we have have a bit better `layer` caching when both of the files are built in same job.

datastore/data-loader/client_knmi_station_ingest.py
fjugipe: Renamed the function and moved `"core"` logic to `main()` in both `data-loaders` https://github.com/EURODEO/e-soh/pull/209/commits/96526fe0f1ebcd43ce1b23784ddbfbaf32240191

datastore/data-loader/client_knmi_station_ingest.py
fjugipe: They have the same performance and it does the same as `functools.partial` in one line. `Starmap` is flexible, but maybe `partial` could be a bit more readable and concise, since the `url` stays the same. 

So refactored to use `functools.partial` in https://github.com/EURODEO/e-soh/pull/209/commits/603ffef4e63f0e5b53b0061d7d7f3ba7f7aa5c41

datastore/data-loader/client_knmi_station.py
fjugipe: Yes, this is something I would have liked to avoid, but it lessens the burden to maintain two different `data-loaders` so that they can be used with same comparison data.

datastore/data-loader/client_knmi_station.py
fjugipe: Yes, that way both of the `data-loaders` have the same `"output"` and we can use the same test comparison data.

datastore/data-loader/client_knmi_station.py
fjugipe: Yes, it is defined as `creator_type: Optional[Literal["person", "group", "institution", "position"]]` in the `model`.

datastore/integration-test/test_knmi.py
fjugipe: `client_knmi_station.py` is inserting `cloud_base_altitude:2.0:point:PT0S` and `cloud_area_fraction:2.0:point:PT0S` twice to the system from the same station through `grpc`. This cannot happen with `ingest`, since `ingest` overrides duplicates. 

So even if the `api` doesn’t return those observations twice with the KNMI testdata, they are still there and are included in the `GetObservations` response. So if we don’t want to use `len` in the test, the other way would be to alter the KNMI `netcdf` files to not have duplicate observations on that station and if we go that route, I think we should take duplicates out of every station to be consistent.

justfile
fjugipe: Yeah, I just made a `Dockerfile` for the tests and the `just` commands.

datastore/integration-test/test_knmi.py
lukas-phaf: I think it is not so much that the there are duplicate observations in the netcdf, but that several observations map to the same "parameter_name". Isn't the best way to solve this to remove the parameters `nc1, nc2, nc3, hc1, hc2 and hc3` from `data-loader/parameters.py`?

.github/workflows/ci.yml
```diff
@@ -61,7 +61,7 @@ jobs:
         run: just services
 
       - name: Load the data into the database
-        run: just load
+        run: just ingest_load
```
lukas-phaf, line 64: The gRPC ingest is not used in the workflow, correct?
How de we ensure it keeps working?
Or is this part of the workflow refactor?

justfile
```diff
@@ -92,6 +95,8 @@ integration:
 unit:
     docker compose run --rm api-unit
 
+ingest_unit:
```
lukas-phaf, line 98: Can we use `ingest-unit` and `ingest-load` as just commands for consistency?

docker-compose.yml
```diff
@@ -136,6 +144,20 @@ services:
       store:
         condition: service_healthy
 
+  ingest_loader:
```
lukas-phaf, line 147: `ingest-loader` for consistency with the rest of the file?

Also, can you remove the first line (the version bit) from the docker-compose? It keeps complaining about this deprecation.

datastore/integration-test/test_knmi.py
fjugipe: These went completely under my radar. Decided to drop `nc` and `hc` so I don't need to alter the test data and took the `set` out from this test: https://github.com/EURODEO/e-soh/pull/209/commits/086e592c27bdf770029569d1cf44e5fe40908a0a

.github/workflows/ci.yml
fjugipe: I would prefer to separate `on push` tests to test only the service changed. So changes to for example `ingest` would launch `ingest-unit` and `ingest-load` with `integration`. Changes to `api` would launch `unit`, `load` and `integration`. `on pull_request`. But that is part of the workflow refactor.

datastore/integration-test/test_knmi.py
lukas-phaf: I am surprised this works. The `nc1`, `nc2` and `nc3` still insert data in the same spot, right?

.github/workflows/ci.yml
lukas-phaf: Ok, let's get back to this with the workflow refactor.

datastore/integration-test/test_knmi.py
fjugipe: Yes, they go to the exact same spot.

The problem was because `nc1`, `nc2` and `nc3` have the `first`, `second` and `third layer` in their _long_name_ (_example: `nc1` = Cloud Amount First Layer_). The `layers` have no real meaning, but they turn to levels of `0.0`, `1.0` and `2.0` as a previous hack: 
```
# We want "level" to be numeric, so get rid of "first", "second" and "third".
# Note that this level has no meaning, it is just a hack for this test dataset
    if level == "first":
        level = "0"
    if level == "second":
        level = "1.0"
    if level == "third":
        level = "2.0"
 ```

The `nc` is _Total cloud cover_ by its _long_name_ and since it doesn't contain any level information, it used the "default" level of `2.0` with `generate_parameter_name` function, so it clashed with `nc3`. Same applies to `hc` and `hc3`

Comment (lukas-phaf) [reply datastore/integration-test/test_knmi.py]: Ah, thanks for clarifying.

Can we just drop nc1,2,3 and hc1,2,3 and get rid of all the level hacks for those? And keep nc and hc instead? Or is that a lot of refactoring in the output files? It would make it easier to understand when doing maintenance in the future.

Comment (fjugipe) [reply datastore/integration-test/test_knmi.py]: Those are only used in a few output files, so we can do that quite easily. 

I will do some other adjustments to the `parameters.py` also, instead of:
```
# Drop parameters that are not CF compliant
                if not standard_name or standard_name in ["precipitation_duration", "rainfall_duration"]:
                    continue
```

fjugipe: Done in: https://github.com/EURODEO/e-soh/pull/209/commits/fef62dffc7612ad5370d1c0da1e432fec773c033

lukas-phaf (approved):

PR #210, opened 2024-10-23
Bufr ingest fixes

istvans-mn:
Fix invalid measure period.
Reduce log noise: skip BUFR subset,  if the measure is too late or too early.

### DIFF
--- a/datastore/datastore/common/common.go
+++ b/datastore/datastore/common/common.go
@@ -75,7 +75,7 @@ func initValidTimeRange() {
 	loTimeName := "LOTIME"
 	hiTimeName := "HITIME"
 	defaultLoTimeSecs := int64(86400)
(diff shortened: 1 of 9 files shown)


datastore/datastore/common/common.go
```diff
@@ -75,7 +75,7 @@ func initValidTimeRange() {
 	loTimeName := "LOTIME"
 	hiTimeName := "HITIME"
 	defaultLoTimeSecs := int64(86400)
-	defaultHiTimeSecs := int64(-2) // add leeway to reduce risk of missing the newest observations
+	defaultHiTimeSecs := int64(-600) // the reference time of visual observations is the next hour
```
Teddy-1000, line 78: This change makes it possible to ingest data that have observations time up to 10 minutes in to the future?

Teddy-1000 (approved):

PR #211, opened 2024-10-23
Add Monitoring to test local performance improvements

Jeffrey-Vervoort-KNMI:
When running just local you can access the monitoring.

- E.g. in Grafana you can view statistics about the gRPC methods
- E.g. in Grafana you can view statistics about the most expensive and most called database queries.


grafana/provisioning/datasources/datasource.yaml
```diff
@@ -0,0 +1,29 @@
+apiVersion: 1
```
Jeffrey-Vervoort-KNMI, line 1: Feedback from the review:

- Look into if it is possible to show pending requests. Thus requests that are still in flight and need to be processed.
- Look into if the titles are consistent with the type. E.g. if the graph has a type which can change (us, ms, seconds), do not mention it in the title or table header.

jo-asplin-met-no:
The gRPC Server dashboard in Grafana has the same three panel groups (or whatever you call them) for each of three jobs. The panel groups are gRPC General, gRPC Methods, and Go, while the jobs are prometheus, prometheus-postgres-exporter, and store. Does this mean that prometheus and prometheus-postgres-exporter are also written in Go and use gRPC for communication?

datastore/database/README.md
```diff
@@ -0,0 +1,10 @@
+# Database logs
+Set UID and GID of the postgres user
+`sudo chown -R 101:103 ./logs`
```
jo-asplin-met-no, line 3: Where/when is this `logs` directory created? I can't find it on my machine (after running `just local test`).

Jeffrey-Vervoort-KNMI: This is no longer applicable. I forgot to remove this readme. It is now removed.

Jeffrey-Vervoort-KNMI:
> The gRPC Server dashboard in Grafana has the same three panel groups (or whatever you call them) for each of three jobs. The panel groups are _gRPC General_, _gRPC Methods_, and _Go_, while the jobs are _prometheus_, _prometheus-postgres-exporter_, and _store_. 

I think the Grafana term is `variables`. At this moment, except for the interval, you always want to keep them at their default value. The other values are not that interesting. For example, database names is a dynamic list retrieved from Prometheus. It consists of all the databases in the Postgres container, namely our database and the default databases. I did not mind it, but for clarity, it could be better to only have `data` as value. What do you think?

> Does this mean that prometheus and prometheus-postgres-exporter are also written in Go and use gRPC for communication?

They are written in Go. I don't think they use gRPC. I made the dashboard this way as I thought for our most common use case `store` everything is in one place. If we want to view the Go performance of "prometheus" or the "exporter" we can. But as I said I expect we seldom look at it, so I think there is something to say to change the variable to only `store`. What do you think?

jo-asplin-met-no:
To avoid confusion I would personally have preferred to move the two monitoring servers (prometheus and exporter) out in a separate dashboard (typically called Monitoring Infrastructure or something like that). Or even delete them altogheter, but that's not important I think. It's interesting to have them there.

grafana/provisioning/datasources/datasource.yaml
jo-asplin-met-no: A note regarding _pending requests_: this seems to be almost the same as _Number of active connections_ in the _Database Performance_ dashboard, but not quite. For example, a given client call to a gRPC method (= a "request") could involve multiple database queries. Moreover, the number of ongoing (incomplete) query executions and the number of active connections isn't necessarily a one-to-one relationship (depending on the definition of "active" I guess).

Teddy-1000: I don't think the number of active connections are very important in gRPC. All the clients connected to the gRPC server should reuse their connection as long as possible and only re-establish it if it drops, which is 1 per API process. The number rpc calls in flight is a lot more useful.

Jeffrey-Vervoort-KNMI: Looking into the metric again. I gave it a poor name. I renamed it to in flight requests. The metric represents the unary rpc requests which are in flight.

I will no longer implement the number of active connections as Amund is right, those aren´t very important.

PR #214, opened 2024-11-19
Issue 178 levels filtering with combination

### LINKED ISSUE #178: [API] Allow multiple ranges with levels filtering
**User should be able to use the custom dimension for `levels` with multiple ranges.** 

For example `/collections/observations/locations?levels=1.0/2.0,8.0/10.0` should be allowed to get all observations between 1.0m and 2.0m and also between 8.0m and 10.0m 

**User should be able to use custom dimension filtering for `levels` with ranges and values together.**

For example `/collections/observations/locations?levels=0.0/1.5,2.5,10.0` should be allowed to get all observations between 0.0m and 1.5m, but also observations from 2.5m and 10.0m

fjugipe:
Redo of #181 

Allow using multiple _ranges_ and multiple _repeating-intervals_ when filtering data with `levels` custom dimension. 
Using for example: 
- `/collections/observations/locations?levels=1.0/2.0, 8.0/10.0`
- `/collections/observations/locations?levels=0.0/1.5, 2.5, 10.0`

is possible after the merge.

/closes #178

### DIFF
--- a/api/utilities.py
+++ b/api/utilities.py
@@ -5,6 +5,7 @@ import isodate
 from datetime import datetime
 from datetime import timedelta
 from typing import Tuple
+from itertools import chain
 
 from isodate import ISO8601Error
 import datastore_pb2 as dstore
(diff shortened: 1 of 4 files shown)


api/utilities.py
```diff
@@ -181,25 +182,46 @@ def get_periods_or_range(periods: str) -> list[str]:
...
+                get_z_values_from_interval(level_or_range)
+                if len(level_or_range) > 2
+                else (
+                    get_z_values_from_range(level_or_range)
+                    if len(level_or_range) == 2
+                    else [convert_m_to_cm(level_or_range[0])]
+                )
+            )
+            for level_or_range in values
+        ]
```
lukas-phaf, line 207: This is way beyond my tolerance for list (in)comprehension.

Can we just call the old `get_z_levels_or_range` (the rewrite to `get_z_values_from_range` is good),
and just call it with something like:
```
values = itertools.chain(*[get_z_levels_or_range(lever_or_range) for level_or_range in split_and_strip(z)])
```
I haven't tested this, but you probably get the idea.

What do you think?

fjugipe: Made the old `get_z_levels_or_range` to a nested function so it would be clearer: https://github.com/EURODEO/e-soh/pull/214/commits/f119e88e7f8435f72a3718fd7b372bd3e8fc2121

PR #216, opened 2024-11-29
Add configureable root path

Teddy-1000:
Add option to configure the FAST API root path with the `FASTAPT_ROOT_PATH` env variable.

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -1,6 +1,7 @@
 # Run with:
 # For developing:    uvicorn main:app --reload
 import logging
+import os
 
 import metadata_endpoints
 from brotli_asgi import BrotliMiddleware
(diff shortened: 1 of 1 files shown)


fjugipe (approved):

api/main.py
```diff
@@ -30,7 +31,10 @@ def setup_logging():
 
 logger = logging.getLogger(__name__)
 
-app = FastAPI(swagger_ui_parameters={"tryItOutEnabled": True})
+app = FastAPI(
+    swagger_ui_parameters={"tryItOutEnabled": True},
+    root_path=os.getenv("FASTAPI_ROOT_PATH", ""),
```
lukas-phaf, line 36: If I set a root path of `/123`, I now get responses at both:
```
http://localhost:8008/123/collections
http://localhost:8008/collections
```

Looking at the [docs](https://fastapi.tiangolo.com/advanced/behind-a-proxy/), I am not sure this is what `root_path` is intended for (and probably explains the strange behavior).

Teddy-1000: I think is is expected behavior as explained by the docs. When this is set, you are expected to route through the proxy.

PR #223, opened 2025-02-05
Add wis2 compliant mqtt publishing

Teddy-1000:


docker-compose.yml
```diff
@@ -105,10 +105,17 @@ services:
     environment:
       - DSHOST=${DSHOST:-store}
       - DSPORT=${DSPORT:-50050}
-      - MQTT_HOST=${MQTT_HOST}
+      - MQTT_HOST=${MQTT_HOST:-mqtt}
       - MQTT_USERNAME=${MQTT_USERNAME}
```
shamlymajeed, line 109: - MQTT_USERNAME=${MQTT_USERNAME:-}
- MQTT_PASSWORD=${MQTT_PASSWORD:-}

Teddy-1000: The difference here is that without the `:-` the variable is set to an empty string if the varible is unset in the launching env. If we add `:-` they will not be set at all in the container if unset in the launching env. I need them to be set to something.

shamlymajeed: oh .ok!

shamlymajeed (approved):

PR #224, opened 2025-02-14
Change base_url handling in API

lukas-phaf:
Remove code with misleading comments which does not achieve anything.

**This should be double-checked on the test system to make sure it works with the proxy configuration.**


Teddy-1000 (approved):
Confirmed correct behind proxy.

PR #231, opened 2025-04-07
Properly support different database names, and support SSL

teunoknmi:
- Allow to use a different database name in the migration container with backwards compatability.
- Update docs
- Might require https://github.com/EUMETNET/e-soh/pull/230 because of updates to latest in scripts.


fjugipe (approved):
works locally.

PR #239, opened 2025-05-13
Split up coverage publication so it can work with PR's from fork

lukas-phaf:
This is based on this example:
https://github.com/MishaKav/pytest-coverage-comment/pull/153

This should work, but can only be tested when it is merged, because:
`This event will only trigger a workflow run if the workflow file exists on the default branch.`
See: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_run

### DIFF
--- a/.github/workflows/publish-coverage.yml
+++ b/.github/workflows/publish-coverage.yml
@@ -1,0 +1,63 @@
+name: Comment coverage report on the pull request
+on:
+  workflow_run:
+    workflows: ["CI"]
+    types:
+      - completed
+
+jobs:
+  post-coverage-report:
+    runs-on: ubuntu-latest
+    if: >
+      github.event.workflow_run.event == 'pull_request' &&
+      github.event.workflow_run.conclusion == 'success'
+    steps:
+      - name: 'Download artifacts'
+        uses: actions/github-script@v7
+        with:
+          script: |
+            var artifacts = await github.rest.actions.listWorkflowRunArtifacts({
+               owner: context.repo.owner,
+               repo: context.repo.repo,
+               run_id: ${{github.event.workflow_run.id }},
+            });
+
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-coverage.yml
```diff
@@ -0,0 +1,63 @@
...
+    workflows: ["CI"]
+    types:
+      - completed
+
+jobs:
+  post-coverage-report:
+    runs-on: ubuntu-latest
+    if: >
+      github.event.workflow_run.event == 'pull_request' &&
+      github.event.workflow_run.conclusion == 'success'
```
fjugipe, line 13: The `CI` workflow is also triggered by _on push_ events and previously we had coverage commenting to push comments. This workflow change will stop doing that, was that the intention?

 I mean I'm fine with the coverage posted only on `pull requests`, since the `on push` comments at this point don't really add any value.

fjugipe (approved):
Looks a nice way to accomplish this without compromising `write` permissions to forks.

.github/workflows/publish-coverage.yml
lukas-phaf: Didn't this statement exclude it before?
`if: github.event_name != 'push' || github.event.ref_type != 'tag'`

Also, where would it push it?

fjugipe: The combined logic of `if: github.event_name != 'push' || github.event.ref_type != 'tag'` translates to:

The job is going to run if:
- the event is not a `push` i.e. it is a `pull_request`
- the event is a `push`, but the reference type is **NOT** a _tag_

It was made to prevent commenting when a tag was created for a release.

Push comments are in the commit comments for example [595e1c2](https://github.com/EUMETNET/e-soh/commit/595e1c233a110e4fed33bb2bd5d9b45dc893fa49#commitcomment-156916356)

PR #245, opened 2025-05-16
(WIP) Bulk insert

lukas-phaf:


datastore/datastore/storagebackend/postgresql/putobservations.go
```diff
@@ -214,71 +325,118 @@ func getObsTime(obsMdata *datastore.ObsMetadata) (*timestamppb.Timestamp, error)
...
+	)
+   , ins AS (
+		INSERT INTO geo_point (point)
+			SELECT * FROM input_rows
+			ON CONFLICT (point) DO NOTHING
+			RETURNING id, point
+	)
+	SELECT id, ST_X(point::geometry), ST_Y(point::geometry) FROM ins
+	UNION
+	SELECT c.id, ST_X(c.point::geometry), ST_Y(c.point::geometry) FROM input_rows
```
lukas-phaf, line 384: Use `postgis.PointS` instead of `ST_X/Y`. Also see #249.

datastore/datastore/storagebackend/postgresql/putobservations.go
```diff
@@ -125,67 +128,173 @@ func getTSColValsUnique(colName2Val map[string]interface{}) ([]interface{}, erro
...
+						%s    -- actual values
+				) t (%s) OFFSET 1
+			) t ORDER BY %s   -- ORDER BY for consistent order to avoid deadlocks
+		)
+		, ins AS (
+			INSERT INTO time_series (%s)
+				SELECT * FROM input_rows
+				ON CONFLICT ON CONSTRAINT unique_main
+					DO UPDATE SET %s  -- do update of fields not in unique constraint
+						WHERE %s      -- only if at least one value is actually different, to avoid table trashing
```
jo-asplin-met-no, line 188: I guess you mean t**h**rashing? (three occurrences of that typo in the change). But anyway, why is the concept of thrashing relevant here? My understanding of thrashing is that a system suddenly performs poorly because some data structure no longer fits fully in physical memory, thus resulting in expensive I/O for paging to disk. I guess this can potentially happen here if we see a lot of "memory leaked" rows every time there's a conflict/match in the original upsert operation. But do we know for sure that that's what Postgres will resort to? I'm sure you have already thought about this, so I just wanted to check. We should only use the term thrashing if that's actually what can happen. But maybe the term "table thrashing" means something different? (tried to google a bit ...)

Also (while I'm at it), the statement "to avoid table thrashing" suggests that this code now guarantees that thrashing (whatever that means) can't happen. But is it possible to make such a guarantee? If not, maybe replace with something less categorical, such as "to minimize risk of thrashing"? And again: since I'm not entirely sure what you mean by thrashing, I'm mostly guessing here (and possibly may appear more annoying than helpful ;-) ).

lukas-phaf: Postgres does writes a new row every time we hit this statement without the `WHERE` clause, even if nothing has changed. Which means every time we insert an observation for this timeseries. From the link in point 1:
```
Most importantly, with [PostgreSQL's MVCC model](https://www.postgresql.org/docs/current/mvcc-intro.html) UPDATE writes a new row version for every target row, no matter whether the row data changed. This incurs a performance penalty for the UPSERT itself, table bloat, index bloat, performance penalty for subsequent operations on the table, VACUUM cost. A minor effect for few duplicates, but massive for mostly dupes.
```
I also observed this: write 10 observations, and look at the time-series table. Rewrite the same observations, the time-series table now has entries in a different order (this was before the 'ORDER BY` on line 181 was added), a clear indication that the row was rewritten.

So, most of the rows in the time-series table will be rewritten thousands of time.

I summarized all the bad issues in the comment above by "table trashing". I actually didn't know "thrashing" is a different word... but know that I do, still feel "trashing" is pretty appropriate... i.e. it makes a mess of the table.

lukas-phaf: I am happy to use another word, though... but no make it less categorical, because what was happening before (without the `WHERE`) is was pretty bad.

datastore/datastore/storagebackend/postgresql/putobservations.go
```diff
@@ -125,67 +128,173 @@ func getTSColValsUnique(colName2Val map[string]interface{}) ([]interface{}, erro
...
+			log.Printf("db.Query() failed: %v", err)
+			if e, ok := err.(*pq.Error); ok {
+				if len(e.Detail) > 0 {
+					log.Printf("db.Query() failed: DETAIL: %v", e.Detail)
+				}
+				if len(e.Hint) > 0 {
+					log.Printf("db.Query() failed: HINT: %v", e.Hint)
+				}
+			}
+			return nil, fmt.Errorf("tx.Query() failed: %v", err)
```
jo-asplin-met-no, line 267: typo: s/tx.Query/db.Query

datastore/datastore/storagebackend/postgresql/putobservations.go
```diff
@@ -125,67 +128,173 @@ func getTSColValsUnique(colName2Val map[string]interface{}) ([]interface{}, erro
...
+		// at the start of this transaction.
+		// A simple solution is to just rerun the query.
+		// See under "Concurrency issue 1" for a similar case here: https://stackoverflow.com/a/42217872
+		if len(tsIDmap) == len(mapTScolVals) {
+			return tsIDmap, nil
+		}
+		log.Printf("In upsertTSs(): concurrency issue detected: 'len(tsIDmap)=%v', 'len(mapTScolVals)=%v', "+
+			"retrying db query...", len(tsIDmap), len(mapTScolVals))
+	}
+	return nil, fmt.Errorf("upsertTSs() failed: still concurrency issues afer 3 retries")
```
jo-asplin-met-no, line 297: So what do we do if we end up in this situation?

datastore/datastore/storagebackend/postgresql/putobservations.go
```diff
@@ -214,71 +323,118 @@ func getObsTime(obsMdata *datastore.ObsMetadata) (*timestamppb.Timestamp, error)
...
+		var p postgis.PointS
+		for rows.Next() {
+			rows.Scan(&id, &p)
+			gpIDmap[GeoPoint{p.X, p.Y}] = id
+		}
 
-	return id, nil
+		// Under concurrent load, if another process is adding the same entry, in which case this transaction
+		// waited for it to complete. Once completed, this transaction would not change it (because of the WHERE),
+		// and therefore not return the row. The SELECT would also not return it, because it see the snapshot
```
jo-asplin-met-no, line 411: typo: s/see/sees

PR #249, opened 2025-05-19
Issue 107 - add GetLocations

jo-asplin-met-no:
Fixes [Issue 107](https://github.com/EUMETNET/e-soh/issues/107).


datastore/datastore/storagebackend/postgresql/getlocations.go
```diff
@@ -0,0 +1,143 @@
...
+	"google.golang.org/grpc/codes"
+)
+
+// getLocs gets into locs the locations of the most recent observation of the distinct platforms
+// that match request and tspec.
+//
+// Returns nil upon success, otherwise error.
+func getLocs(
+	db *sql.DB, request *datastore.GetLocsRequest, tspec common.TemporalSpec,
+	locs *[]*datastore.LocMetadata) error {
```
lukas-phaf, line 21: Why not return this as the return value of the function?

datastore/datastore/storagebackend/postgresql/getlocations.go
```diff
@@ -0,0 +1,143 @@
...
+		err := createObsQueryVals(
+		request.GetSpatialPolygon(), request.GetSpatialCircle(), request.GetFilter(), tspec,
+		&phVals)
+	if err != nil {
+		return fmt.Errorf("createLocsQueryVals() failed: %v", err)
+	}
+
+	// define and execute query
+	query := fmt.Sprintf(`
+		SELECT DISTINCT ON (platform, parameter_name)
```
lukas-phaf, line 35: Just so I understand it correctly, the query get the latest `obstime_instant` (in the requested time window) for each `platform, parameter_name` pair, because of the combination of:
```
DISTINCT ON (platform, parameter_name)
```
and
```
ORDER BY platform, parameter_name, obstime_instant DESC
```

Neat trick!

protobuf/datastore.proto
```diff
@@ -323,3 +326,27 @@ message GetExtentsResponse {
   TimeInterval temporal_extent = 3 [json_name = "temporal_extent"];
   BoundingBox spatial_extent = 4 [json_name = "spatial_extent"];
 }
+
+//---------------------------------------------------------------------------
+
+message GetLocsRequest {
+  // NOTE: the below fields have the same meaning as in GetObsRequest
+  bool temporal_latest = 1;
```
lukas-phaf, line 334: This parameter doesn't actually do anything, correct?

datastore/datastore/storagebackend/postgresql/getlocations.go
```diff
@@ -0,0 +1,143 @@
...
+		err = rows.Scan(&point, &platform, &platformName, &paramName)
+		if err != nil {
+			return fmt.Errorf("rows.Scan() failed: %v", err)
+		}
+
+		if platform != currPlatform { // next platform, possibly the first one!
+			if currPlatform != "" { // add for previous platform
+				if len(currPoints) == 0 {
+					return fmt.Errorf("programming error [1]: len(currPoints) == 0")
+				}
```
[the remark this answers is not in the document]
jo-asplin-met-no: Might a good idea. I'll look into it.

protobuf/datastore.proto
jo-asplin-met-no: You're right. I'll remove it.

datastore/datastore/storagebackend/postgresql/getlocations.go
```diff
@@ -0,0 +1,143 @@
...
+		SELECT DISTINCT ON (platform, parameter_name)
+			point,
+			platform,
+			platform_name,
+			parameter_name
+		FROM observation
+		JOIN time_series on observation.ts_id = time_series.id
+		JOIN geo_point ON observation.geo_point_id = geo_point.id
+		WHERE %s AND %s AND %s AND %s
+		ORDER BY platform, parameter_name, obstime_instant DESC
```
lukas-phaf, line 44: Can you not get postgres to do all the work, including the point sorting? Something like:
```
SELECT platform, platform_name, parameter_name, ST_X(point) AS px, ST_Y(point) AS py FROM 
   (SELECT DISTINCT ON (platform, parameter_name)
	point, platform, platform_name, parameter_name
    ...
    ORDER BY platform, parameter_name, obstime_instant DESC
    ) t ORDER BY platform, parameter_name, py, px LIMIT 1
```

The syntax is probably wrong...

datastore/datastore/storagebackend/postgresql/getlocations.go
```diff
@@ -0,0 +1,143 @@
...
+	if err != nil {
+		return fmt.Errorf("createLocsQueryVals() failed: %v", err)
+	}
+
+	// define and execute query
+	query := fmt.Sprintf(`
+		SELECT DISTINCT ON (platform, parameter_name)
+			point,
+			platform,
+			platform_name,
```
lukas-phaf, line 38: If a `platform` happens to have multiple `platform_name`'s, we take the "first" one, correct? This is probably fine, as you have to pick one.

But do we want to `ORDER BY` on that as well, to get a "stable" result?

datastore/datastore/storagebackend/postgresql/getlocations.go
lukas-phaf, line 44: For future reference, it is also possible to do the collection in postgres:
```
SELECT min(point::geometry), platform, array_agg(distinct platform_name) as platform_names,
       array_agg(distinct parameter_name) as parameter_names FROM
                (SELECT DISTINCT ON (platform, parameter_name)
                   point,
                   platform,
                   platform_name,
                   parameter_name
               FROM observation
                        JOIN time_series on observation.ts_id = time_series.id
                        JOIN geo_point ON observation.geo_point_id = geo_point.id
               WHERE ....
               ORDER BY platform, parameter_name, obstime_instant DESC) t GROUP by platform
```
I here take the `min` of the points (whatever that may mean). It is also possible to get a bbox using `ST_Extent`.

datastore/datastore/storagebackend/postgresql/getlocations.go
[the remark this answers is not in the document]
jo-asplin-met-no: Done!

datastore/datastore/storagebackend/postgresql/getlocations.go
jo-asplin-met-no: Noted!

datastore/datastore/storagebackend/postgresql/getlocations.go
jo-asplin-met-no: Your wish is my command! (in this case anyway ;-) )

datastore/datastore/storagebackend/postgresql/getlocations.go
```diff
@@ -0,0 +1,147 @@
...
+
+	rows, err := db.Query(query, phVals...)
+	if err != nil {
+		return nil, fmt.Errorf("db.Query() failed: %v", err)
+	}
+	defer rows.Close()
+
+	// process rows ...
+
+	addResultItem := func(point postgis.PointS, pform, pformName string, paramNames *[]string) {
```
lukas-phaf, line 63: What does defining this function add over just putting the code in the one place it is used?

Also, why use a nested function here?

datastore/datastore/storagebackend/postgresql/getlocations.go
```diff
@@ -0,0 +1,147 @@
...
+				Lon: point.X,
+				Lat: point.Y,
+			},
+			Platform:       pform,
+			PlatformName:   pformName,
+			ParameterNames: *paramNames,
+		})
+	}
+
+	getRepresentativePoint := func(points *[]postgis.PointS) postgis.PointS {
```
lukas-phaf, line 75: Same questions as for `addResultItem`

datastore/datastore/storagebackend/postgresql/getlocations.go
```diff
@@ -0,0 +1,147 @@
...
+			return (*points)[i].X < (*points)[j].X // ... and secondarily on longitude
+		})
+		return (*points)[0]
+	}
+
+	// per platform info
+	type pformInfo struct {
+		platformName string
+		points       *[]postgis.PointS
+		paramNames   *[]string
```
lukas-phaf, line 89: Why do you use a pointer to slice in the struct, instead of a slice (without the pointer)?

datastore/datastore/storagebackend/postgresql/getlocations.go
```diff
@@ -0,0 +1,147 @@
...
+
+	// populate pformInfos from rows
+	for rows.Next() {
+
+		var (
+			point        postgis.PointS
+			platform     string
+			platformName sql.NullString
+			paramName    string
+		)
```
lukas-phaf, line 102: Take these out of the `for` loop?

datastore/datastore/storagebackend/postgresql/getlocations.go
```diff
@@ -0,0 +1,147 @@
...
+				Lat: point.Y,
+			},
+			Platform:       pform,
+			PlatformName:   pformName,
+			ParameterNames: *paramNames,
+		})
+	}
+
+	getRepresentativePoint := func(points *[]postgis.PointS) postgis.PointS {
+		sort.Slice(*points, func(i, j int) bool {
```
lukas-phaf, line 76: It feels inefficient to do a full sort to get the min value.
Use `slice.MinFunc` instead?
https://pkg.go.dev/slices#MinFunc

datastore/datastore/storagebackend/postgresql/getlocations.go
jo-asplin-met-no: To have deep copying.

datastore/datastore/storagebackend/postgresql/getlocations.go
jo-asplin-met-no: Good point - will do!

datastore/datastore/storagebackend/postgresql/getlocations.go
lukas-phaf: You mean deep sharing? Or does go automatically copy it because it is a pointer?

I was asking specifically here because it is just a temporary struct that doesn't get copied around, and the whole struct is only used through a pointer:
`pformInfos := map[string]*pformInfo{}`

So effectively these slices now have pointers to pointers...

jo-asplin-met-no: I don't know. It works. Feel free to change it later.

datastore/datastore/storagebackend/postgresql/getlocations.go
jo-asplin-met-no: done

fjugipe:
# API CHANGES
Changes made for the `API` to use new `GetLocsRequest`

## Refactors
- Generate `QUDT` dict based on `std_name` units in system resulting in smaller dict
- Use `GetLocsRequest` together with `GetTSAGRequest` to generate `<collection>/locations` _FeatureCollection_

## Misc
- New `just` command to copy _std_unit_names.json_ to required locations

api/generate_qudt_units.py
```diff
@@ -26,20 +26,38 @@ def generate_qudt_dictionary():
...
+            "type": qudt_units.get(value["unit"], {}).get("type", "https://qudt.org/schema/qudt/UNKNOWN"),
+            "value": qudt_units.get(value["unit"], {}).get("value", "-"),
+        }
     return qudt_dict
 
 
 def main():
-    qudt_dict = generate_qudt_dictionary()
+    qudt_units = get_qudt_units()
+    qudt_dict = generate_qudt_dictionary(qudt_units, "std_name_units.json")
```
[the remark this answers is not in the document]
fjugipe: I thought about doing it in the startup, but my reasoning against it was that if for some reason one of the url's would not be accessible, the service would not start. This could become a problem if we need to spin up more services or for some reason the service crashes and is restarted, and there is no one immediately to handle / notice the problems. So in my opinion building it to the image, which results maybe 1mb larger image, and having a reliable startup was a better solution in this case.

I chose to use the top level variable, since it is slightly more performant than a cached function and _"conversion"_ is called all the time. The `dict` is also quite small and doesn't have a lot of key-value pairs or a complex logic behind it so the memory imprint is small.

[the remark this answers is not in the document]
lukas-phaf: Fair enough... another advantage of your approach is that multiple services would be guaranteed to have the same dict...

datastore/datastore/storagebackend/postgresql/getlocations.go
jo-asplin-met-no: I'll move it inline.

api/routers/edr.py
```diff
@@ -108,98 +107,57 @@ async def get_locations(
...
+    uniq_parameters: Set[str] = set()
+    for loc in locations:
+        features.append(
+            Feature(
+                type="Feature",
+                id=loc.platform,
+                properties={
+                    "name": loc.platform_name if loc.platform_name else f"platform-{loc.platform}",
+                    "detail": f"https://oscar.wmo.int/surface/rest/api/search/station?wigosId={loc.platform}",
+                    "parameter-name": list(loc.parameter_names),
```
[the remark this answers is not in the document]
lukas-phaf: That's pretty neat.

I also learned that quicksort can be O(n^2) in this case, depending on the algorithm that is used to choose the pivot... Ouch!

PR #251, opened 2025-05-20
issue-175: Update /collections/obserations/dataset self links.

Jeffrey-Vervoort-KNMI:
https://github.com/EUMETNET/e-soh/issues/175

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -79,6 +79,7 @@ async def get_collections(request: Request) -> Collections:
     base_url = create_url_from_request(request)
     return await metadata_endpoints.get_collections(base_url, all_collections)
 
+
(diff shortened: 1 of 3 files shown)


api/routers/feature.py
```diff
@@ -174,6 +177,47 @@ async def get_dataset_metadata():
                 f"{extent.temporal_extent.end.ToDatetime().strftime('%Y-%m-%dT%H:%M:%SZ')}",
             ],
         ],
+        "links": json.dumps(
```
Teddy-1000, line 180: I kind want this the other way arraound.
I want to have the filter base_url, conformance_url and docs_url.
This way they can used in the jinja2 file instead.

Teddy-1000 (approved):

PR #258, opened 2025-05-22
Issue 254

Jeffrey-Vervoort-KNMI:


api/routers/feature.py
```diff
@@ -130,7 +131,7 @@ async def search_timeseries(
 
     time_series = await get_obs_request(obs_request)
 
-    return formatters.metadata_formatters[f](time_series.observations)
+    return formatters.metadata_formatters[f](time_series.observations, str(request.base_url))
```
Jeffrey-Vervoort-KNMI, line 134: Passing the request.base_url around does not feel clean, but we want to use it in the GeoJSON. What do you think?

Teddy-1000: I don't really see any other options, since this value is only available in the request object.

Teddy-1000 (approved):

PR #259, opened 2025-05-28
Replace query by one that performs better by 30-40%.

lukas-phaf:
When comparing the new GetLocations code with the code on `main`, I noticed that the `/locations` endpoint on `main` used to take about 100-125 ms for the default test dataset (the one you get with `just load`), and with the new code it is actually slower (around 330 ms).

I looked into this, and managed to get the time down to about 200 ms by a simple change to the query, which brings it more in line with the query that datastore used to do for the `/locations` endpoint.

Compared to `main` both the `issue_107` and the `loc_query` have signifcantly lower traffic between the API and the datastore. For example, for the Netatmo data, it goes down from 24MB to 8MB.

So overall, I think this is still an improvement.

### DIFF
--- a/datastore/datastore/storagebackend/postgresql/getlocations.go
+++ b/datastore/datastore/storagebackend/postgresql/getlocations.go
@@ -36,7 +36,7 @@ func getLocs(
 
 	// define and execute query
 	query := fmt.Sprintf(`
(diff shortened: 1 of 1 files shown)


jo-asplin-met-no:
Looks good!

Issue #269, opened 2025-07-10
[FEATURE REQ]start_datetime, end_datetime ingest


istvans-mn:
The ingester rejects the JSON schema if there is a `start_datetime` and `end_datetime` pair instead of a `datetime` field.

PR #270, opened 2025-07-10
Allow to ingest start_datetime, end_ datetime instead of datetime

istvans-mn:
Issue #269 
- Copy end_datetime to datetime
- Generate period from end_datetime - start_datetime
- Ingest schema update
- Properties class validation update


Teddy-1000:
Feature not needed in OASIS.

PR #279, opened 2025-11-06
Set mqtt v5

Teddy-1000:


fjugipe (commented):
I think we should also add  `- MQTT_PROTOCOL_VERSION=${MQTT_PROTOCOL_VERSION}` to `docker-compose.yml`

This would allow controlling the _protocol version_ via `ci/config/env.list`, which is used for example in all the `just` commands

Otherwise there is just a small typo, everything else looks fine.

PR #280, opened 2025-12-09
Rodeo bufr library

istvans-mn:
The BUFR core codes was removed by this PR. The BUFR codes are in the [rodeo-bufr-tools](https://pypi.org/project/rodeo-bufr-tools/) package.

### DIFF
--- a/ingest/requirements.in
+++ b/ingest/requirements.in
@@ -16,5 +16,5 @@ python-multipart~=0.0.9
 requests == 2.31.0
 isodate~=0.6.1
 prometheus-fastapi-instrumentator
-rodeo-bufr-tools ~= 0.3.2
+rodeo-bufr-tools ~= 0.3.3
 python-dateutil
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -108,12 +108,14 @@ services:
     environment:
       - DSHOST=${DSHOST:-store}
       - DSPORT=${DSPORT:-50050}
+      - DYNAMICTIME=$DYNAMICTIME
+      - LOTIME=$LOTIME
+      - HITIME=$HITIME
```
Teddy-1000: Are these only used in the bufr library?
I think it would be strange if messages posted to the JSON endpoint would get a different error then the bufr messages. 
Or is this just for filtering out observations that are outside of the time window, leaving the valid ones?

istvans-mn: Yes, it is only for for filtering, it skips observations if the datetime is out of time window.

Teddy-1000 (approved):

PR #298, opened 2026-03-09
261: /locations should return all parameters available for a station, even when filtering.

Jeffrey-Vervoort-KNMI:
https://github.com/EUMETNET/e-soh/issues/261


datastore/datastore/storagebackend/postgresql/getlocations.go
```diff
@@ -36,21 +36,33 @@ func getLocs(
...
+				AND %s
+			ORDER BY ts_id, platform, obstime_instant DESC
+		) t ORDER BY platform, parameter_name;
 		`,
 		timeFilter,
 		geoFilter,
 		int64MdataFilter,
-		stringMdataFilter)
+		stringMdataFilter,
+		timeFilter,
```
Jeffrey-Vervoort-KNMI: It works, but adding timeFilter twice does not look clean. An alternative like using templates, e.g. https://gobyexample.com/text-templates increases the number of lines, and because of that does also not feel perfect. Any opinion on this?

datastore/datastore/storagebackend/postgresql/postgresql.go
```diff
@@ -321,7 +321,7 @@ func addWhereCondMatchAnyPatternForInt64(
 			whereExprOR = append(whereExprOR, "TRUE")
 		} else { // fall back to regular string matching
 			index++
-			expr := fmt.Sprintf("(lower(%s::text) LIKE lower($%d))", colName, index)
+			expr := fmt.Sprintf("(%s::text ILIKE $%d)", colName, index)
```
lukas-phaf: This is also in #295, make sure we don't get it duplicate.

datastore/datastore/storagebackend/postgresql/getlocations.go
```diff
@@ -36,21 +36,33 @@ func getLocs(
 
 	// define and execute query
 	query := fmt.Sprintf(`
```
jo-asplin-met-no, line 38: At first glance I feel that the new query is perhaps more complex than necessary. **Note:** this is not based on an actual understanding of what it does, only the fact that there are now redundant parts etc. Would it be possible to simplify it?
I am going to take a closer look, though. Maybe I find that this is a good solution after all.

jo-asplin-met-no: After spending a bit more time staring at the statement, I now think it makes better sense and is not too complex after all.

Jeffrey-Vervoort-KNMI: I agree that it looks more complex. It is a trade-off between simplicity and performance.

The queries performance are the same for small databases, but when having a large number of rows in the tables, it outperforms the simpler query. Which makes it valuable for our production environment.

How it works: It uses CTE to split up the work efficiently, it applies filters early so it has a smaller set when joining.

jo-asplin-met-no: Sounds reasonable. But I guess my next question then is why _platform_ is treated in a special way here? In the [original issue](https://github.com/EUMETNET/e-soh/issues/261), the problem was that you only got one _parameter_?

I mean, intuitively I would expect instead `WITH parameter_names AS (SELECT DISTINCT ON (parameter_name) parameter_name ...` and so on.

But I might be wrong.

Jeffrey-Vervoort-KNMI: I will try to explain it based on my understanding. I think because the Postgres query planner does most of the work already. When I put other parts in the with statements, I did not saw any changes in performance or execution order. So I think it already picked the best path for the execution of parameter_names. At least when I made variations on Lukas his query  and compared them. But for platform it did make a difference, the execution order changed after putting it in a CTE, and the query became more efficient when having a large number of rows.

jo-asplin-met-no: Still sounds reasonable I guess. Looks good to me!

jo-asplin-met-no:
Looks good!

jo-asplin-met-no (approved):

PR #305, opened 2026-03-17
Issue_288_add_alt_platforms

jo-asplin-met-no:
Add alt_platforms as a time series metadata field for keeping alternative (WIGOS) identifiers for stations.


ingest/api/model.py
```diff
@@ -221,9 +221,16 @@ class Properties(BaseModel):
         ),
     )
     platform_name: Optional[str] = Field(None, description=("Human readable name for the platform."))
+    alt_platforms: Optional[List[str]] = Field(
+        ...,
```
istvans-mn:
```suggestion
        None,
```
... means required.

ingest/api/model.py
```diff
@@ -359,14 +366,24 @@ def check_datetime_iso(self) -> "Properties":
...
+                assert (
+                    b.isdigit() and 0 <= int(b) <= 65534
+                ), f"Block {j} in input '{name}' is not a valid integer in range [0,65534]: '{wid}'"
+                j = j + 1
+
+            assert 0 < len(blocks[-1]) <= 16, f"In input '{name}', last block of WIGOS is too long: '{wid}'"
+
+        val_wigos_id("platform", self.platform)
+        i = 0
+        for ap in self.alt_platforms:
```
istvans-mn: Check None:
```
if self.alt_platforms:
```

jo-asplin-met-no: Done

ingest/api/model.py
jo-asplin-met-no: Done

